PR #1, opened 2017-11-13
PlatformBootManagerLib requires CapsuleLib  library

kalyan-nagabhirava:
"Instance of library class [CapsuleLib] is not found"  build error is coming for Hikey platform, to resolve this issue
added CapsuleLib to "LibraryClasses.common" section


nate-desimone:
Hi @kalyan-nagabhirava,

The tianocore project does not accept pull requests. Instead, we use a code contribution process that is very similar to the Linux kernel patch submission process. For details, please see here: https://github.com/tianocore/edk2/blob/master/Readme.md#code-contributions

In a nutshell, send your patch to devel@edk2.groups.io using git send-email and make sure you read and can abide by the rules of the developer certificate of origin and BSD licensing requirements.

Hope that helps. If you have any questions or need some help drop us a note on the mailing list and we would be happy to assist!

PR #8, opened 2020-01-16
Silicon/Broadcom/Net: Add Genet stub driver to setup MAC

pbatard:
The RPi4 has a Gigabit Broadcom Genet platform device. In order to
assist in booting the platform in an ACPI environment, it is
desirable if the hardware can describe itself rather than passing
parameters via DSDT/DSD. One way to achieve this is to assure that
the adapter is preprogrammed with the correct ethernet MAC address
which, on the Raspberry Pi, can be achieved by using a dedicated
firmware interface call.

The driver is added under Silicon on account that Genet NICs are
not specific to the Raspberry Pi and could be extended for other
platforms. This is also why the dedicated PcdBcmGenetRpiMacInit
PCD is introduced for the RPi specific MAC initialization.

Signed-off-by: Pete Batard <pete@akeo.ie>

### DIFF
--- a/Platform/RaspberryPi/RPi4/RPi4.dsc
+++ b/Platform/RaspberryPi/RPi4/RPi4.dsc
@@ -346,6 +346,9 @@
 
   gEfiMdeModulePkgTokenSpaceGuid.PcdFirmwareVersionString|L"EDK2-DEV"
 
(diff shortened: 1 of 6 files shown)


pbatard:
Wrong repo. Please disregard.

PR #11, opened 2020-05-21
readme: update links to Linaro toolchains

dkalowsk:
The old path of "toolchain/binaries/latest" does not resolve to an active page for download.  Linaro has broken the "toolchain/binaries/latest" path into the latest per version of GCC (i.e. "toolchain/binaries/latest-4" and "toolchain/binaries/latest-6").  This change believes that anyone wanting to use the Linaro toolchain will be wanting to use the most recent GCC version as well.

Signed-off-by: Dan Kalowsky <dank@deadmime.org>

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -57,8 +57,8 @@ X64                 | x86_64-linux-gnu-
 
 ### GCC
 Linaro provides GCC toolchains for
-[aarch64-linux-gnu](https://releases.linaro.org/components/toolchain/binaries/latest/aarch64-linux-gnu/)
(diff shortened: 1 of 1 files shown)


bcran:
This change is now obsolete due to my recent commit to link to the Arm GNU-A toolchain page in https://github.com/tianocore/edk2-platforms/commit/7bf73ecc3c470d7d80e047ac93e7fc3c0ab23499 .

hrw:
Closing due to @bcran comment.

PR #16, opened 2021-01-06
spell check, section #Booting the firmware

Redpillbug:
corrected uSD card to SD card , please correct me if such a thing called uSD card exists, but I think it meant SD card


JohnAZoidberg:
uSD card looks like μSD card, which means Micro SD card.

bcran:
Agreed, uSD is likely referring to a micro SD card.

PR #21, opened 2021-03-10
Reset DWC2 on all raspberry pi boards

rtreffer:
Setting the attributes to 0 causes `DwHcReset` to never execute any code.
Setting the attributes to `EFI_USB_HC_RESET_HOST_CONTROLLER` will result in a full USB controller reset.

This will not change anything for boards that initialize the USB controller in the firmware (RPI3/4). But some boards (CM4 and potentially PI400) do not initialize the port early and leave it unconfigured / powered of.

Tested on my CM4 with the official IO board.


rtreffer:
Posted to edk2 devel instead.

PR #24, opened 2021-04-02
Platform/ARM/VExpressPkg: Fix unused but set

RaMdsC:
Remove unused but set variables in GetArmNameSpaceObject. These caused a
build crash due to -Werror=unused-but-set-variable.

Signed-off-by: Adrián Herrera <adr.her.arc.95@gmail.com>


hrw:
Both variables are used in current code.

PR #25, opened 2021-05-18
Enable SecureBoot flag build on U540 and bootable on RISCV QEMU

joshloo:
This patch enables build with -DSECURE_BOOT_ENABLE=TRUE flag.
However, some library included are null libraries here, and during boot,
there's some unsupported function during SecureBootConfigDxe,
specifically InvalidateInstructionCacheRange that needs to be fixed

Signed-off-by: Loo Tung Lun <joshlootunglun@gmail.com>

### DIFF
--- a/Platform/SiFive/U5SeriesPkg/FreedomU540HiFiveUnleashedBoard/U540.dsc
+++ b/Platform/SiFive/U5SeriesPkg/FreedomU540HiFiveUnleashedBoard/U540.dsc
@@ -118,6 +118,9 @@
   DebugPrintErrorLevelLib|MdePkg/Library/BaseDebugPrintErrorLevelLib/BaseDebugPrintErrorLevelLib.inf
 
 !if $(SECURE_BOOT_ENABLE) == TRUE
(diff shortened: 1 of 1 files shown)


Platform/SiFive/U5SeriesPkg/FreedomU540HiFiveUnleashedBoard/U540.dsc
```diff
@@ -419,6 +422,7 @@
     <LibraryClasses>
       NULL|SecurityPkg/Library/DxeImageVerificationLib/DxeImageVerificationLib.inf
   }
+  SecurityPkg/VariableAuthenticated/SecureBootConfigDxe/SecureBootConfigDxe.inf
```
JohnAZoidberg, line 425: This is already present at Line 531.
But it would probably be nice to get it closer together. I.e. put it here and remove it from below

bcran:
The TianoCore project doesn't use Pull Requests (yet), so you'll need to send the patch to devel@edk2.groups.io (see https://edk2.groups.io) to have it considered for inclusion.

hrw:
rebase please

PR #50, opened 2022-11-15
BoardBdsHookLib [Patch V2] GCC compiler fix + PCD based Uefi shell load

abdattar:
Patch 1: handles GCC compiler boot time page fault exception by making copy of PcdPtr values.
Patch 2: Adds PCDs to hold UEFI shell file Guid and description


abdattar:
submitting V3 version

PR #99, opened 2023-09-26
U5SeriesPkg/FreedomU540HiFiveUnleashedBoard: Support MMU with SV39/48…

Xxfore:
By following https://github.com/riscv-admin/riscv-uefi-edk2-docs to build edk2 for riscv qemu(u540).
`build -a RISCV64 -p Platform/SiFive/U5SeriesPkg/FreedomU540HiFiveUnleashedBoard/U540.dsc -t GCC5`
it will occur build error like :
image

It will build success after add missing library class. 
image


hrw:
Already fixed in 72efe4b6a848c2a8062297f142d94439aadf39ca

PR #108, opened 2023-11-03
CI check for StarFive platform

johnchewyy:
CI check for StarFive platform


davidlt:
I assume the name for this PR is wrong. This looks like StarFive JH7110 port. Is anyone looking at this?

PR #115, opened 2023-11-30
UserAuthFeaturePkg/UserAuthenticationSmm: Support Standalone MM.

xuweiintel:
Refactor UserAuthenticationSmm to support Standalone MM.
- Factor out variable lock code logic that references boot services.
- UserAuthenticationStandaloneMmDxe is added to lock the variables. It is only used for UserAuthenticationStandaloneMm.
- UserAuthenticationStandaloneMm doesn't lock the variables, needs to rely on UserAuthenticationStandaloneMmDxe to do the lock.
- UserAuthenticationSmm still locks the variables by itself, no need to include UserAuthenticationStandaloneMmDxe.
- Register gEfiEventExitBootServicesGuid notify which is used by the StandaloneMmCore.

Since gEdkiiVariableLockProtocolGuid is a deprecated interface, use gEdkiiVariablePolicyProtocolGuid to lock password variables instead.

Cc: Dandan Bi <dandan.bi@intel.com>
Cc: Nate DeSimone <nathaniel.l.desimone@intel.com>
Cc: Liming Gao <gaoliming@byosoft.com.cn>

### DIFF
--- a/Features/Intel/UserInterface/UserAuthFeaturePkg/Include/UserAuthFeature.dsc
+++ b/Features/Intel/UserInterface/UserAuthFeaturePkg/Include/UserAuthFeature.dsc
@@ -6,7 +6,7 @@
 # INF files to generate AutoGen.c and AutoGen.h files
 # for the build infrastructure.
 #
-# Copyright (c) 2020 - 2021, Intel Corporation. All rights reserved.<BR>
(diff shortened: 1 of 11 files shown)


hrw:
merged in bd1c9ed97dc96a53671786ea73102c78da4c7e32

PR #133, opened 2024-03-29
Changes about MADT table in MinPlatformPkg AcpiPlatform driver

td36:


Platform/Intel/MinPlatformPkg/Acpi/AcpiTables/AcpiPlatform.c
```diff
@@ -18,6 +18,8 @@ typedef struct {
   UINT32   Flags;
   UINT32   SocketNum;
   UINT32   Thread;
+  UINT8    CoreType;
+  BOOLEAN  IsBsp;
```
niruiyu, line 22: Is "IsBsp" needed? Can we use "Flags"?

Platform/Intel/MinPlatformPkg/Acpi/AcpiTables/AcpiPlatform.c
```diff
@@ -197,6 +297,11 @@ CreateCpuLocalApicInTable (
     }
   }
 
+  //
+  // Perform quick sort
+  //
+  QuickSort (CpuApicIdOrderTable, mNumberOfCpus, sizeof (EFI_CPU_ID_ORDER_MAP), (BASE_SORT_COMPARE)ApicIdOrderTableCompare, &SortBuffer);
```
niruiyu, line 303: The ApicIdOrderTableCompare () is too complex.
How about?
1. do an insert-sort to move big cores in front of small cores.
2. Move BSP in the beginning.

Platform/Intel/MinPlatformPkg/Acpi/AcpiTables/AcpiPlatform.c
td36: IsBsp was dropped. Also I kept the original Flags usage since it will be assigned to the Flags in LOCAL_X2APIC_STRUCTURE directly.

Just use a local variable BspIndex to record this info.

Platform/Intel/MinPlatformPkg/Acpi/AcpiTables/AcpiPlatform.c
td36: Acctually I moved BSP in the beginning first and then perform the insert-sort for the rest cores.
If perform step1 first, the Bsp may be moved to another position. Another for-loop is needed to find where BSP is.

niruiyu (approved):

PR #151, opened 2024-06-05
Ampere/JadePkg: Add secure boot default keys initialization

bcran:
This allows to initialize secure boot with the default factory keys embedded in firmware flash image.

For example, to incorporate PK, KEK, and DB default keys, specify the corresponding key files in the Jade.dsc as follows:

DEFINE DEFAULT_KEYS        = TRUE
DEFINE PK_DEFAULT_FILE     = path/to/PK.crt
DEFINE KEK_DEFAULT_FILE1   = path/to/KEK.crt
DEFINE DB_DEFAULT_FILE1    = path/to/DB1.crt
DEFINE DB_DEFAULT_FILE2    = path/to/DB2.crt

Signed-off-by: Nhi Pham <nhi@os.amperecomputing.com>
Reviewed-by: Rebecca Cran <rebecca@os.amperecomputing.com>


nhivp:
The patch was recently pushed to the master branch.

PR #181, opened 2024-08-16
ManageabilityPkg:  ManageabilityTransportSsifLib: resource cleanup fix

ghbaccount:


nhivp:
This looks good to me. Thanks for the fix.

_(I can not see the approve button although I'm able to merge it. @changab please review and merge if it also looks good to you)_

Features/ManageabilityPkg/Library/ManageabilityTransportSsifLib/Dxe/ManageabilityTransportSsif.c
```diff
@@ -362,21 +362,22 @@ ReleaseTransportSession (
...
+    Status              = EFI_SUCCESS;
+    mSingleSessionToken = NULL;
   }
 
   if (SsifTransportToken != NULL) {
     FreePool (SsifTransportToken->Token.Transport->Function.Version1_0);
     FreePool (SsifTransportToken->Token.Transport);
     FreePool (SsifTransportToken);
-    mSingleSessionToken = NULL;
-    Status              = EFI_SUCCESS;
```
changab, line 379: I want to confirm with you is this removal correct?

ghbaccount: Oh. The condition was not changed in this module.
I'll fix it.

This was moved under the condition 
  if (mSingleSessionToken == SsifTransportToken) {
     Status              = EFI_SUCCESS;
     mSingleSessionToken = NULL;
   }
  
Actually I have another version of this changes, but I didn't published it, because it mostly the same, but the patch got bigger. I've added addition check that SsifTransportToken != NULL.
It's a minor issue, because it could be NULL only if sizeof (UINTN) value was passed to function as pointer parameter.

I tried to fix original tricky code with a minimal changes.

ghbaccount:
Please, look at the updated version with the fixed issue mentioned above.
The logic remained the same: the resource passed to the function is freed if the pointer is valid, EFI_SUCCESS is returned only when the condition (mSingleSessionToken == SsifTransportToken) is true.

Features/ManageabilityPkg/Library/ManageabilityTransportSsifLib/Dxe/ManageabilityTransportSsif.c
nhivp: @changab Please review.

changab (approved):

PR #186, opened 2024-09-03
CODEOWNERS: Fix Yi Qian's username

bcran:
Yi Qian is @yqian4, not @yqian.


leiflindholm (approved):

bcran:
I don't have write access to edk2-platforms, so somebody else will need to merge this.

PR #190, opened 2024-09-06
Revert RiscVQemuServerPlatform

vlsunil:
Leif pointed that this platform should not be in master until qemu has this platform support. Also, it is missing Maintainer entry. He suggested to have this in new devel- branch instead which I agree. So, reverting this platform from master. A request to add a new devel branch will be created where this platform will be added.


vlsunil:
@andreiw @ChaiEvan @sfu5 - FYI

vlsunil:
@leiflindholm - Could you please help with review?

leiflindholm (approved):

PR #192, opened 2024-09-10
Platform/RPi4: Allocate more space for UEFI image

jlinton:
The Pi's UEFI image was sized without thought for much future growth, and its become apparent that its largly outgrown the allocated space when various debug options or edk2 featurees are enabled. Since these images tend to exist on disks which are usually in the GB+ ranges doubling the size of the RPI_EFI.fd image shouldn't be a problem.

This change must be accompanied by build edk2-non-osi changes to TFA which update the RPI3_PRELOADED_DTB_BASE=0x3f0000 and config.txt changes as documented in the README being patched.

Since we are adjusting the image size, lets reserve a 64k block for future PCC channel data structures that are shared between TFA and ACPI as well immediately before the DTB.

### DIFF
--- a/Platform/RaspberryPi/RPi4/RPi4.dsc
+++ b/Platform/RaspberryPi/RPi4/RPi4.dsc
@@ -284,6 +284,11 @@
   gEfiMdePkgTokenSpaceGuid.PcdPerformanceLibraryPropertyMask|1
   gEfiMdePkgTokenSpaceGuid.PcdPostCodePropertyMask|0
   gEfiMdePkgTokenSpaceGuid.PcdUefiLibMaxPrintBufferSize|320
+  #
+  # Follows right after the FD image. (bump the size again)
+  #
+  gRaspberryPiTokenSpaceGuid.PcdFdtBaseAddress|0x003e0000
+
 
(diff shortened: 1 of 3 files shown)


jlinton:
I only updated the rpi4, because I've not seen a problem with the rpi3. But this will need to be merged along with a new TFA to edk-non-osi and config.txt updates to PFTF. I can open them if everyone thinks this is worthwhile.

ardbiesheuvel:
This looks reasonable to me - would it make sense to increase the size of the DT allocation too, now that we are at it?

jlinton:
That sounds reasonable, so I've doubled it as well, that provides a bit of breathing space if someone adds a bunch of overlays/etc since the base rpi4-b DTB's are all in the mid 50K range already.

This latest update has needs a bit more through testing to validate nothing got broken, which I'm going to do when I return home in a couple hours.

jlinton:
The 64k DTB appears to require further TFA changes since the 64k size is hardcoded in MAP_NS_DTB

ardbiesheuvel:
You mean code changes as opposed to rebuilding it with different defines on the command line, which we need to do in any case, right?

leiflindholm (approved):

PR #202, opened 2024-09-16
 Platform/ARM/Juno: Use RngDxeLib

pierregondois:
Juno's RngLib implementation is:

- BaseRngLib.inf if a secure RngLib is enforced
- BaseRngLibTimerLib.inf if a non-secure RngLib is tolerated

BaseRngLib.inf relies on the Arm's RNDR instruction. The instruction
returns a DRBG-generated random number. The DRBG used is considered
as secure.
The RNDR instruction is available if FEAT_RNG is set. The Juno doesn't
support it.

When security is enforced (i.e. ENABLE_UNSAFE_RNGLIB is not set),
the Juno cannot generate secure random numbers through the RngLib.
Secure random numbers could be generated by using the Juno's TRNG.
This can be done by:

- using the RngDxeLib implementation of the RngLib
- RngDxeLib relies on the RngDxe
- the RngDxe has access to the TRNG

### DIFF
--- a/Platform/ARM/JunoPkg/ArmJuno.dsc
+++ b/Platform/ARM/JunoPkg/ArmJuno.dsc
@@ -219,7 +219,7 @@
   #
   # Juno Support Trng. Override PcdEnforceSecureRngAlgorithms.
   #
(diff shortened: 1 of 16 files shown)


pierregondois:
@samimujawar

samimujawar (approved):

PR #207, opened 2024-09-20
Addition of PcdStandaloneMmEnable has broken build of BoardPkg(s)

nate-desimone:
REF: https://bugzilla.tianocore.org/show_bug.cgi?id=4853

When the PcdStandaloneMmEnable FeaturePcd was added, it was not added to MinPlatformFeaturesPcd.dsc.inc

Because of this, any existing BoardPkg that does not have Standalone MM support will no longer build because the default value of FALSE is not declared during DSC parsing.

The fix is to add the PCD to MinPlatformFeaturesPcd.dsc.inc

### DIFF
--- a/Platform/Intel/MinPlatformPkg/Include/Dsc/MinPlatformFeaturesPcd.dsc.inc
+++ b/Platform/Intel/MinPlatformPkg/Include/Dsc/MinPlatformFeaturesPcd.dsc.inc
@@ -7,7 +7,7 @@
 # A board can enable a feature by configuring the PCD in the board DSC file
 # after this file has been included.
 #
-# Copyright (C) 2022 Intel Corporation
+# Copyright (C) 2022 - 2024 Intel Corporation
 #
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 #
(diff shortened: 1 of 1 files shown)


ChaselChiu (approved):

Platform/Intel/MinPlatformPkg/Include/Dsc/MinPlatformFeaturesPcd.dsc.inc
```diff
@@ -7,7 +7,7 @@
 # A board can enable a feature by configuring the PCD in the board DSC file
 # after this file has been included.
 #
-# Copyright (C) 2022 Intel Corporation
+# Copyright (C) 2022 - 2024 Intel Corporation
```
SaiChaganty, line 10: Correct me if I am wrong but do we need to update the current year of the file change or just leave the year of file creation?

SaiChaganty (approved):

Platform/Intel/MinPlatformPkg/Include/Dsc/MinPlatformFeaturesPcd.dsc.inc
nate-desimone: For Tianocore it is optional.

PR #219, opened 2024-09-30
Silicon/Ampere: Set EFI_ACPI_6_3_LOW_POWER_S0_IDLE_CAPABLE in FADT

bcran:
Set the EFI_ACPI_6_3_LOW_POWER_S0_IDLE_CAPABLE flag in the FADT.

The ACPI specification says the LOW_POWER_S0_IDLE_CAPABLE flag means that the platform is able to achieve power savings in S0 similar to or better than those typically in S3. It effectively means that the system will not achieve any power savings by transitioning to S3. Since Altra doesn't support the S3 state, set this flag.

This resolves the FWTS S0idleNoFADT error.


chuongtranle (commented):
Reviewed-by: Chuong Tran <chuong@os.amperecomputing.com>

nhivp (approved):

PR #225, opened 2024-10-22
Drivers/OptionRomPkg: Add Renesas uPD720202 firmware loader

bexcran:
Add the Renesas uPD720202 firmware loader based on Ard Biesheuvel's patch at
https://edk2.groups.io/g/devel/topic/patch_optionrompkg_add/30905893#

The uPD72020x USB 3.0 chipset family supports two modes of operation. Either the firmware is stored in an external EEPROM chip and downloaded into the chipset at boot time, or it must be uploaded into the chipset by the operating system / driver.
The second option always works and overrides any firmware stored into the EEPROM of the card.

More technical details are discussed in a blog post at https://mjott.de/blog/881-renesas-usb-3-0-controllers-vs-linux.

### DIFF
--- a/Drivers/OptionRomPkg/RenesasFirmwarePD720202/README.md
+++ b/Drivers/OptionRomPkg/RenesasFirmwarePD720202/README.md
@@ -1,0 +1,51 @@
+# RenesasFirmwarePD720202
+
+RenesasFirmwarePD720202 is a firmware loader for Renesas
+uPD720201/uPD720202 family USB 3.0 controllers.
+
+It provides the functionality to upload the firmware required for
+certain extension cards before they can used.
+
+## Usage
+
+ * First, a firmware image for the chipset is required.
+   The final release 2.0.2.6 (K2026090.mem) can be downloaded from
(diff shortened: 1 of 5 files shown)


Drivers/OptionRomPkg/RenesasFirmwarePD720202/README.md
```diff
@@ -0,0 +1,54 @@
...
+ * The firmware image is uploaded into the chipset.
+   This process is non-persistent, the chipset RAM is cleared when the power supply is removed.
+
+To use this driver, update the platform .dsc file:
+```
+  #
+  # Renesas PD720202 XHCI firmware uploader
+  #
+  Drivers/OptionRomPkg/RenesasFirmwarePD720202/RenesasFirmwarePD720202.inf {
+    <LibraryClasses>
```
leiflindholm, line 24: Hmm...
Doesn't any real platform need to define a global resolution for DxeServicesLib anyway, since it's a dependency for DxeMain?

bexcran: Oh, good point. I'll remove that from the README.

leiflindholm (approved):

PR #233, opened 2024-10-26
Platform/Intel: Add build_bios.py option to skip rebuilding BaseTools

mikebeaton:
# Description

When running build_bios.py manually it is convenient to be able to avoid rebuilding BaseTools (which also involves re-running the BaseTools tests) on every run.

Even though the time to build BaseTools (once they are already built) and re-run the BaseTools tests is relatively small compared to the time taken to fully build any of these platforms, we may be debugging an issue which prevents a given platform from building entirely, or early, and which has nothing to do with changes in BaseTools, in which case it is convenient not to have to wait for the BaseTools tests to complete each time round the debug loop.

- [ ] Breaking change?
- [ ] Impacts security?
- [ ] Includes tests?

## How This Was Tested

Local builds.

## Integration Instructions

N/A

### DIFF
--- a/Platform/Intel/build_bios.py
+++ b/Platform/Intel/build_bios.py
@@ -34,7 +34,7 @@ except ImportError:
     import configparser
 
 
-def pre_build(build_config, build_type="DEBUG", silent=False, toolchain=None):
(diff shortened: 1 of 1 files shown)


bexcran:
This looks reasonable.

nate-desimone (approved):

PR #248, opened 2024-11-18
ARM/SgiPkg: fix build failure on SgiPlatformMm

vbpandya:
commit 8bc44608b8ea ("MdeModulePkg: Add ImagePropertiesRecordLib") makes build fail. This patch adds ImpagePropertiesRecordLib and related libraries to fix the build.


samimujawar:
@vbpandya Thank you for this fix.
The change-ID needs to be dropped, otherwise this patch looks good to me.

vbpandya:
> @vbpandya Thank you for this fix. The change-ID needs to be dropped, otherwise this patch looks good to me.

Thank you for catching that Sami!
Updated accordingly.

PR #722, opened 2025-01-07
Add GitHub Issue Templates and Automation

makubacki:
Brings edk2-platforms GitHub issues into parity with edk2.

---

**.github: Add GitHub issue templates**

REF: https://github.com/tianocore/edk2/discussions/5926

Adds GitHub form isssue templates for bugs, documentation requests,
and feature requests.

These files define the form structure in YAML to be rendered by
GitHub in the edk2-platforms repo issues area.

Due to the large amount of diverse content on edk2-platforms an
attempt was made to classify code areas in the best way for the
current organization of the code path in the tree. Please suggest
any corrections in the PR.

---

**.github: Add issue automation workflows**

REF: https://github.com/tianocore/edk2/discussions/5926

Adds workflows to manage labels on issues based on issue content.

This uses the same flow as already used for issues in the edk2 repo.

Workflows:

- `issue-assignment` - Performs actions when an issue is assigned.
  - Currently, removes the `state:needs-owner` label.
- `issue-triage` - Assigns initial labels to the issue based on data
  entered when the issue was created.
  - The policies for applying labels are defined in
    - `advanced-issue-labeler.yml`
	- Note: Based on https://github.com/marketplace/actions/advanced-issue-labeler
- `scheduled-maintenance` - Runs every hour to perform clean up work
  need on issues.
  - Currently, closes issues that have had the `state:wont-fix` label
    applied.

---

Tested on my edk2-platforms fork.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,20 @@
+## @file
+# GitHub issue configuration file.
+#
+# This file is meant to direct contributors familiar with GitHub's issue tracker
+# to the external resources used by TianoCore.
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+##
+
+contact_links:
+  - name: Reporting Security Issues
+    url: https://github.com/tianocore/tianocore.github.io/wiki/Reporting-Security-Issues
+    about: Read the wiki page that describes the process here
+  - name: EDK II Development Mailing List
+    url: https://edk2.groups.io/g/devel
+    about: Ask questions on the mailing list (devel@edk2.groups.io)
+  - name: EDK II Platforms Discussions
+    url: https://github.com/tianocore/edk2-platforms/discussions
+    about: You can also reach out on the Discussion section for this repository
--- a/.github/advanced-issue-labeler.yml
+++ b/.github/advanced-issue-labeler.yml
@@ -1,0 +1,52 @@
+# Defines the mappings between GitHub issue responses and labels applied to the issue.
+#
+# Copyright (c) Microsoft Corporation.
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# For more information, see:
+# https://github.com/redhat-plumbers-in-action/advanced-issue-labeler
+
+policy:
+  - section:
+
+    # Issue Template - Urgency Dropdown
+    - id: ['urgency']
+      block-list: []
+      label:
+        - name: 'priority:low'
+          keys: ['Low']
+        - name: 'priority:medium'
+          keys: ['Medium']
+        - name: 'priority:high'
+          keys: ['High']
+
+    # Issue Template - Code First
+    - id: ['code_first']
+      block-list: []
+      label:
+        - name: 'type:code-first'
+          keys: ['Yes']
+
+    # Issue Template - Fix Owner Dropdown
+    - id: ['fix_owner', 'request_owner']
+      block-list: []
+      label:
+        - name: 'state:needs-owner'
+          keys: [
+            'Someone else needs to fix it',
+            'Someone else needs to make the change',
+            'Someone else needs to implement the feature'
+            ]
+        - name: 'state:needs-triage'
+          keys: [
+            'Someone else needs to fix it',
+            'Someone else needs to make the change',
+            'Someone else needs to implement the feature'
+            ]
+
+    # Issue Template - Needs Maintainer Feedback Dropdown
+    - id: ['needs_maintainer_feedback']
+      block-list: []
+      label:
+        - name: 'state:needs-maintainer-feedback'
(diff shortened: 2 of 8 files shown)


.github/advanced-issue-labeler.yml
```diff
@@ -0,0 +1,325 @@
...
+      block-list: []
+      label:
+        - name: 'state:needs-maintainer-feedback'
+          keys: ['Maintainer feedback requested']
+
+    # Issue Template - Packages Impacted
+    - id: ['packages_impacted']
+      block-list: []
+      label:
+        - name: 'package:all'
```
mdkinney, line 58: This is a lot of labels.  Is there a better way to indicate the package or platform without adding labels?

leiflindholm: It's also an added maintenance burden.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,20 @@
...
+contact_links:
+  - name: Reporting Security Issues
+    url: https://github.com/tianocore/tianocore.github.io/wiki/Reporting-Security-Issues
+    about: Read the wiki page that describes the process here
+  - name: EDK II Development Mailing List
+    url: https://edk2.groups.io/g/devel
+    about: Ask questions on the mailing list (devel@edk2.groups.io)
+  - name: EDK II Platforms Discussions
+    url: https://github.com/tianocore/edk2-platforms/discussions
+    about: You can also reach out on the Discussion section of this repository
```
leiflindholm, line 20: Apologies for nitpicking on something I assume is copied straight from existing text for edk2, but:
The discussion section is not part of the repository, it is attached to the repository in the github project.

Could we reword this something like "...on the Discussion section for this repository"?

.github/advanced-issue-labeler.yml
makubacki: I agree it's a lot, but that's also a reflection of how diverse the content is in the repo at the moment. The main reason for the approach is consistency with what is done per package in edk2 (using labels). Labels simplify filtering operations not only in the GitHub UI but also GitHub automation as many GitHub actions have built-in support for labels.

If the primary concern is maintaining a manual list of labels in this file (`.github/advanced-issue-labeler.yml`), I could try to use a function to consistently convert the label value to a string as opposed to the lookup list currently used.

Otherwise, we could have an automated workflow prepend the package name to the issue title after it is submitted. That would still allow textual searching though not as clean as label filtering.

In any case, we would still need to maintain the list of issue paths in the form template YAML files.

.github/ISSUE_TEMPLATE/config.yml
makubacki: Updated

.github/advanced-issue-labeler.yml
mdkinney: I would prefer a string in the title or a tag in a comment with this information so it can be searched.  We can have scripts that verify valid tags/strings, and meetings like the TianoCore BugScrub can review issues that do not have matching tags/strings to get them updated appropriately.

I recommend back propagating this approach to edk2 repo issues as well to reduce number of labels in edk2 repo.

makubacki: There are thousands of issues in edk2 currently assigned against the labels that would need modified. The number of labels is also much less there. That will be disruptive and a separate set of work as it's modifying a large number of existing issues versus setting up the flow for new issues here.

I will update this PR within the next couple of days to add the package path in the issue title.

mdkinney: I agree edk2 is a different task.

The problem with using the title is that if there are multiple tags the title gets very long.

A standard <TagName>:<Value> string can support multiple tags

makubacki: Changes made for this:

1. Removed the label list from this file (`advanced-issue-labeler.yml`).
2. Added a step to the issue submission flow to append the impacted packages.
   - I'd like to add the packages to an area externally visible to the issue as it helps with searching and determining relevancy when looking at a list of issues.
   - To reduce the impact on the title length, the packages are not added if more than 3 are selected. I'm expecting this (or the "All" option) will be the common case making adding the package to the title both brief and common.
   - In any case, the list of actual packages selected will be present in the issue body.

makubacki:
@mdkinney & @leiflindholm, I've made updates per feedback. Please review again when you have time.

mdkinney:
It looks like the current set of labels in edk2-platforms repo has packages from the edk2 repo.  Will those get removed as well?

mdkinney (approved):

makubacki:
> It looks like the current set of labels in edk2-platforms repo has packages from the edk2 repo. Will those get removed as well?

A few issues had those packages associated with them. I updated the issues and removed all edk2-specific labels from this repo.

mdkinney:
It looks like there are a few package labels for edk2-platforms that also need to be converted to the new approach in this PR

PR #728, opened 2025-01-20
Platform/PlatformStandaloneMmPkg/: Disable shadow copying of BFV in StMM

LeviYeoReum:
On PlatformStandaloneMmPkg, the StandaloneMM Boot Firmware Volume (BFV) is not in the Flash area. Instead it is loaded in
to the RAM by optee's code.
Therefore, disable the shadow copying of the BFV in StandaloneMM.

### DIFF
--- a/Platform/StandaloneMm/PlatformStandaloneMmPkg/PlatformStandaloneMmRpmb.dsc
+++ b/Platform/StandaloneMm/PlatformStandaloneMmPkg/PlatformStandaloneMmRpmb.dsc
@@ -113,6 +113,12 @@
 
   gArmTokenSpaceGuid.PcdFfaLibConduitSmc|FALSE
 
+  # The BFV is not located in the Flash area but is loaded in the RAM
+  # by optee's stmm_sp.c instead, therefore no shadow copy is needed.
+  # So disable shadow copy of boot firmware volume while loading StMM drivers.
+  #
+  gStandaloneMmPkgTokenSpaceGuid.PcdShadowBfv|FALSE
+
 [PcdsPatchableInModule]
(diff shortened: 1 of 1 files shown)


samimujawar:
@apalos Can you confirm if this change works in your setup, please?

apalos:
> @apalos Can you confirm if this change works in your setup, please?

Ah i missed this. There's a readme in OPTEE can you update that then?

LeviYeoReum:
> > @apalos Can you confirm if this change works in your setup, please?
> 
> Ah i missed this. There's a readme in OPTEE can you update that then?

I think It doesn't need to update EDK2 Build instructions (in https://optee.readthedocs.io/en/latest/building/efi_vars/stmm.html#edk2-build-instructions)

Because, This patch makes default as disabling BFV shadow copy.

Could you make me know what things should be updated?

apalos:
> > > @apalos Can you confirm if this change works in your setup, please?
> > 
> > 
> > Ah i missed this. There's a readme in OPTEE can you update that then?
> 
> I think It doesn't need to update EDK2 Build instructions (in https://optee.readthedocs.io/en/latest/building/efi_vars/stmm.html#edk2-build-instructions)
> 
> Because, This patch makes default as disabling BFV shadow copy.
> 
> Could you make me know what things should be updated?

That's weird, the default compilation wasn't working until I doubled optees heap size. I'll test again in case that patch wasn't pulled

apalos (approved):

apalos:
Acked-by: Ilias Apalodimas <ilias.apalodimas@linaro.org>
Tested-by: Ilias Apalodimas <ilias.apalodimas@linaro.org>

PR #729, opened 2025-01-20
AmdPlatformPkg: Add support to update ACPI table list

abdattar:
Add support to update ACPI tables list.
So that during boot paltform can add/remove ACPI table.


abdattar:
Hi @changab  and @pbgrimes , could you please the review the code changes.

PR #730, opened 2025-01-21
Reserve ecam v express

vbpandya:
Platform/Arm/VExpressPkg: Reserve the ECAM area in ACPI with RES0 device

Add a RES0 device to the SSDT to reserve the PCI ECAM area.This fixes the firmware bug reported by Linux:

`acpi PNP0A08:00: [Firmware Bug]: ECAM area [mem 0x40000000-0x4fffffff]not reserved in ACPI namespace`

With this change, the above message is replaced with:

`acpi PNP0A08:00: ECAM area [mem 0x40000000-0x4fffffff] reserved by PNP0C02:00`

This patch is based on commit 1e5be97660c6 ("Platform/ARM/JunoPkg:Reserve the ECAM area in ACPI with RES0 device")

REF:https://developer.arm.com/documentation/100964/1127/Base-Platform/Base-Platform-memory/Base-Platform-memory-map?lang=en

Also use same PCD for Juno and VExpress.

### DIFF
--- a/Platform/ARM/VExpressPkg/ArmVExpressPkg.dec
+++ b/Platform/ARM/VExpressPkg/ArmVExpressPkg.dec
@@ -46,6 +46,9 @@
   gArmVExpressTokenSpaceGuid.PcdHdLcdVideoModeOscId|0|UINT32|0x00000005
   gArmVExpressTokenSpaceGuid.PcdArmMaliDpMaxMode|0|UINT32|0x00000008
 
+  gArmTokenSpaceGuid.PcdPciConfigurationSpaceSize|0x10000000|UINT64|0x00000011
+  gArmTokenSpaceGuid.PcdPciConfigurationSpaceLimit|0x4FFFFFFF|UINT64|0x00000012
+
   #
   # Device path of block device on which Fastboot will flash partitions
   #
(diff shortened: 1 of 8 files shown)


vbpandya:
@samimujawar can you please take a look at this PR ?

Platform/ARM/VExpressPkg/ArmVExpressPkg.dec
```diff
@@ -46,6 +46,9 @@
   gArmVExpressTokenSpaceGuid.PcdHdLcdVideoModeOscId|0|UINT32|0x00000005
   gArmVExpressTokenSpaceGuid.PcdArmMaliDpMaxMode|0|UINT32|0x00000008
 
+  gArmTokenSpaceGuid.PcdPciConfigurationSpaceSize|0x10000000|UINT64|0x00000011
```
samimujawar, line 49: How is the gEfiMdePkgTokenSpaceGuid.PcdPciExpressBaseSize different from gArmTokenSpaceGuid.PcdPciConfigurationSpaceSize, can you check, please?

See https://github.com/tianocore/edk2/blob/master/MdePkg/MdePkg.dec#L2538

samimujawar:
@vbpandya any update on the status of this PR?

vbpandya:
> @vbpandya any update on the status of this PR?

This one slipped through the cracks, in the mean time FVP has moved to Dynamic table (makes implementing your feedback easy) which Juno is still on static tables, will close this PR for now, since it will be good to move Juno to Dynamic Table and then use common PCD for both platforms.

samimujawar:
I think we should drop the static tables support and keep Dynamic Tables Framework based support only for Juno. Can you send a separate PR for that, please?

Also, I think the SSDT PCIe generator may be taking care of the Firmware bug issue which the Linux kernel reports. Can you check and report back, please? In that case this PR can be closed.

vbpandya:
With [this](https://github.com/tianocore/edk2-platforms/pull/781) PR merged, this PR is no more required.

PR #744, opened 2025-02-21
Fix hii opcode handle leak

ghbaccount:
Actually this was a part of findings fixed by https://github.com/tianocore/edk2/pull/6424
But when I tried to build /Ampere/JadePkg/Jade.dsc  I've got a bunch of other  [-Werror=maybe-uninitialized] warnings.

### DIFF
--- a/Platform/Ampere/JadePkg/Drivers/SmbiosPlatformDxe/SmbiosPlatformDxe.c
+++ b/Platform/Ampere/JadePkg/Drivers/SmbiosPlatformDxe/SmbiosPlatformDxe.c
@@ -445,6 +445,8 @@ SmbiosPlatformDxeRestoreHiiDefaultString (
   UINT8       Index;
   EFI_STATUS  Status;
 
+  Status = EFI_INVALID_PARAMETER;
+
   if (StrToken == NULL) {
     DEBUG ((
       DEBUG_ERROR,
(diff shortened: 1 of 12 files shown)


nhivp:
Thanks for the patches. Please wait until https://github.com/tianocore/edk2-platforms/pull/260 is merged for fixing the current compilation error with Jade platform.

Platform/Ampere/JadePkg/Drivers/SmbiosPlatformDxe/SmbiosPlatformDxe.c
```diff
@@ -445,6 +445,8 @@ SmbiosPlatformDxeRestoreHiiDefaultString (
   UINT8       Index;
   EFI_STATUS  Status;
 
+  Status = EFI_INVALID_PARAMETER;
```
nhivp, line 448: Should return EFI_SUCCESS at the end of the function instead of returning INVALID_PARAMETER.

Please help check for other changes under Platform/Ampere/JadePkg/Drivers/SmbiosPlatformDxe

ghbaccount: For other files EFI_INVALID_PARAMETER is returned if no entries were added.
This means SmbiosPlatformDxeAddRecord wasn't called because we did not enter the loop, and the comment to functions states
`
  @retval EFI_SUCCESS                The SMBIOS Table was successfully added.
`
So if nothing was added, EFI_SUCCESS should not be returned.

That was the idea of this patch.

nhivp: Thanks for clarifying further the compiler warn. However, I see all error paths are already covered and exits appropriately, the `Status` value at the end of the function should be EFI_SUCCESS.

nhivp: I was concerning about returning the error could stop processing subsequent SMBIOS tables, but it was not. I am fine with the changes now. Thanks for your patience.

nhivp (approved):

PR #756, opened 2025-03-20
Silicon/Marvell: enable clang and NOOPT builds for OdysseyPkg

leiflindholm:
While test building an unrelated change, I noticed I couldn't build the NOOPT target with clang on OdysseyPkg. These changes make that possible.


ndhillonm (approved):
Thank you.

PR #759, opened 2025-03-31
Pg/serial port length

pierregondois:
Add information about serial port length/ access size. This information is used by the ConfigurationManager entities to generate ACPI tables.


samimujawar (approved):
@pierregondois Thank you for these updates.

PR #800, opened 2025-06-16
Platform/ARM/VExpressPkg: Generate ACPI PPTT table

sarah-walker-arm:
Add the required processor hierarchy and private resource structures to allow PPTT to be generated.

### DIFF
--- a/Platform/ARM/VExpressPkg/ConfigurationManager/ConfigurationManagerDxe/ConfigurationManager.c
+++ b/Platform/ARM/VExpressPkg/ConfigurationManager/ConfigurationManagerDxe/ConfigurationManager.c
@@ -84,6 +84,13 @@ EDKII_PLATFORM_REPOSITORY_INFO VExpressPlatRepositoryInfo = {
       CREATE_STD_ACPI_TABLE_GEN_ID (EStdAcpiTableIdSsdtCpuTopology),
       NULL
     },
+    // PPTT Table
+    {
+      EFI_ACPI_6_3_PROCESSOR_PROPERTIES_TOPOLOGY_TABLE_STRUCTURE_SIGNATURE,
+      EFI_ACPI_6_3_PROCESSOR_PROPERTIES_TOPOLOGY_TABLE_REVISION,
+      CREATE_STD_ACPI_TABLE_GEN_ID (EStdAcpiTableIdPptt),
+      NULL
+    },
     // Note: The last 3 tables in this list are for FVP RevC only.
     // IORT Table - FVP RevC
     {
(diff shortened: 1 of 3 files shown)


Platform/ARM/VExpressPkg/ConfigurationManager/ConfigurationManagerDxe/ConfigurationManager.c
```diff
@@ -455,6 +462,446 @@ EDKII_PLATFORM_REPOSITORY_INFO VExpressPlatRepositoryInfo = {
...
+      REFERENCE_TOKEN (Cluster1CoreResources),
+      // CM_OBJECT_TOKEN  LpiToken
+      REFERENCE_TOKEN (CoresLpiRef)
+    }
+  },
+
+
+  // Cache information
+  {
+    // L2 cache
```
pierregondois: Just to be sure, I think a more 'standard' topology would be to have a L3 cache associated to the Package. FWIU, the FVP doesn't have a real topology, so it might be better to create this L3 cache.

Platform/ARM/VExpressPkg/ConfigurationManager/ConfigurationManagerDxe/ConfigurationManager.c
```diff
@@ -455,6 +462,446 @@ EDKII_PLATFORM_REPOSITORY_INFO VExpressPlatRepositoryInfo = {
...
+    {
+      // CM_OBJECT_TOKEN  Token
+      REFERENCE_TOKEN (ProcHierarchyInfo[0]),
+      // UINT32  Flags
+      PROC_NODE_FLAGS (
+        EFI_ACPI_6_3_PPTT_PACKAGE_PHYSICAL,
+        EFI_ACPI_6_3_PPTT_PROCESSOR_ID_VALID,
+        EFI_ACPI_6_3_PPTT_PROCESSOR_IS_NOT_THREAD,
+        EFI_ACPI_6_3_PPTT_NODE_IS_NOT_LEAF,
+        EFI_ACPI_6_3_PPTT_IMPLEMENTATION_NOT_IDENTICAL
```
pierregondois: The Juno is big-little, but the FVP should have the same type of CPUs for both clusters

Platform/ARM/VExpressPkg/ConfigurationManager/ConfigurationManagerDxe/ConfigurationManager.c
```diff
@@ -455,6 +462,446 @@ EDKII_PLATFORM_REPOSITORY_INFO VExpressPlatRepositoryInfo = {
...
+    { REFERENCE_TOKEN (CacheInfo[3]) }  // -> cluster 2's L2 cache
+  },
+  // Resources private to each individual cluster 1 core instance
+  {
+    { REFERENCE_TOKEN (CacheInfo[4]) }, // -> cluster 1 core's L1 I-cache
+    { REFERENCE_TOKEN (CacheInfo[5]) }  // -> cluster 1 core's L1 D-cache
+  },
+
+
+  // Low Power Idle state information (LPI) for all cores/clusters
```
pierregondois, line 846: I think you will also need to add an `_OSC` method to the Dsdt.asl file, cf. `Platform/ARM/JunoPkg/ConfigurationManager/ConfigurationManagerDxe/AslTables/Dsdt.asl`. The `_OSC` method from the Juno seems correct for the FVP and it worked fine for me.

Without it, the OS won't be aware that the platform supports `coordinated low power idle states` (and idle states won't appear under `/sys/devices/system/cpu/cpuX/cpuidle/stateX`)

Cf. ACPI spec:
```
Table 6.13: Platform-Wide _OSC Capabilities DWORD 2
OS Initiated Low Power Idle Support

Note: As part of the handshake provided through _OSC, the OS will pass in flags to indicate whether it supports
Platform Coordinated Low Power Idle or OS Initiated Low Power Idle or both (see Section 8.4.3.2), through flags 7
and 8. The platform will indicate which of the modes it supports in its response by clearing flags that are not supported.
If both are supported, the default is platform coordinated and OSPM can switch the platform to OS Initiated via a
processor architecture specific mechanism. By setting either flag 7 or 8 or both, the OSPM is asserting it supports
any objects associated with Low Power Idle states
```

Platform/ARM/VExpressPkg/ConfigurationManager/ConfigurationManagerDxe/ConfigurationManager.c
samimujawar: I think it may confuse people if we try to depict a L3 cache when it is not actually present in the system. For example we would not add a L3 cache on Juno as it is not present.

sarah-walker-arm: That depends on what exactly this platform is targeting. As Pierre says above, FVP doesn't have a real cache topology. A real Versatile Express would, but the exact configuration would depend on what CPU tile is in use.

pierregondois: Just to share, the SLC should not be represented in the PPTT but in the HMAT table:
https://github.com/tianocore/edk2-platforms/commit/df5e094ef347d12f48c306268c674df40a1357da

So effectively, it might be better to let L1/L2 caches as it is right now

sarah-walker-arm: So do you want me to remove the L3 cache from this change?

pierregondois: Sorry I didn't see you updated the PR.
I just remembered there was a discussion about the difference between a L3 cache and a SLC. The difference being that apparently, the SLC should not be described in the PPTT as it is more a system resource (and not a CPU topology resource).
If this is a L3 cache it should be good.

pierregondois (approved):
Tested on a FVP

Platform/ARM/VExpressPkg/ConfigurationManager/ConfigurationManagerDxe/ConfigurationManager.c
samimujawar: It appears that the FVP Rev C has an option to specify the L2 cache. So keeping the L3 cache should be good.

Platform/ARM/VExpressPkg/ConfigurationManager/ConfigurationManagerDxe/ConfigurationManager.c
```diff
@@ -455,6 +462,464 @@ EDKII_PLATFORM_REPOSITORY_INFO VExpressPlatRepositoryInfo = {
...
+      REFERENCE_TOKEN (Cluster1CoreResources),
+      // CM_OBJECT_TOKEN  LpiToken
+      REFERENCE_TOKEN (CoresLpiRef)
+    }
+  },
+
+
+  // Cache information
+  {
+    // L2 cache
```
samimujawar: Can we update the comment to say L3 cache?

PR #805, opened 2025-06-27
Platform/ARM/VExpressPkg: change PcdDebugPropertyMask for StandaloneMm

LeviYeoReum:
The current PcdDebugPropertyMask setting for StandaloneMm is 0xff,
    which includes the DEBUG_PROPERTY_ASSERT_BREAKPOINT_ENABLED (0x10) flag.
    However, the svc #dbdb instruction is not implemented by the SPMCs.
    
    So in some cases even when ASSERT() fails, execution continues
    instead of halting. For instance, when SPMC_AT_EL3 is used,
    the svc #dbdb instruction is translated into an SMC call by SPMC.
    Since there is no SMC handler for #dbdb, CpuBreakPoint() returns
    SMC_UNKNOWN, resulting in the system continuing execution rather
    than stopping.
    
    In other cases, the svc #dbdb results in an exception in the SPMC,
    e.g. Rust-SPMC.
    
    By changing PcdDebugPropertyMask, let StandaloneMm call CpuDeadloop()
    when it hits an ASSERT().
    
### DIFF
--- a/Platform/ARM/VExpressPkg/PlatformStandaloneMm.dsc
+++ b/Platform/ARM/VExpressPkg/PlatformStandaloneMm.dsc
@@ -120,8 +120,13 @@
 #
 ################################################################################
 [PcdsFixedAtBuild]
+!if $(TARGET) == RELEASE
+  gEfiMdePkgTokenSpaceGuid.PcdDebugPropertyMask|0x21
+!else
+  gEfiMdePkgTokenSpaceGuid.PcdDebugPropertyMask|0x2f
+!endif
+
   gEfiMdePkgTokenSpaceGuid.PcdDebugPrintErrorLevel|0x8000008F
-  gEfiMdePkgTokenSpaceGuid.PcdDebugPropertyMask|0xff
   gEfiMdePkgTokenSpaceGuid.PcdDebugClearMemoryValue|0xAF
 
   ## PL011 - Serial Terminal.


Platform/ARM/VExpressPkg/PlatformStandaloneMm.dsc
```diff
@@ -120,8 +120,13 @@
 #
 ################################################################################
 [PcdsFixedAtBuild]
+!if $(TARGET) == RELEASE
+  gEfiMdePkgTokenSpaceGuid.PcdDebugPropertyMask|0x21
+!else
+  gEfiMdePkgTokenSpaceGuid.PcdDebugPropertyMask|0x2f
```
mgamal, line 126: Is there any reason why it's not 0xef instead of 0x2f since we only need to toggle the 5th bit (DEBUG_PROPERTY_ASSERT_BREAKPOINT_ENABLED)? I see these values are used in other dsc files, so it might make sense to use them, but I just wonder why?

(Note: The value is 0xff is also used in Juno, we might need to modify it there too).

LeviYeoReum: Hi @mgamal.

> Is there any reason why it's not 0xef instead of 0x2f since we only need to toggle the 5th bit (DEBUG_PROPERTY_ASSERT_BREAKPOINT_ENABLED)? I see these values are used in other dsc files, so it might make sense to use them, but I just wonder why?



If you see the definition flag for this:

```
//
// Declare bits for PcdDebugPropertyMask
//
#define DEBUG_PROPERTY_DEBUG_ASSERT_ENABLED       0x01
#define DEBUG_PROPERTY_DEBUG_PRINT_ENABLED        0x02
#define DEBUG_PROPERTY_DEBUG_CODE_ENABLED         0x04
#define DEBUG_PROPERTY_CLEAR_MEMORY_ENABLED       0x08
#define DEBUG_PROPERTY_ASSERT_BREAKPOINT_ENABLED  0x10
#define DEBUG_PROPERTY_ASSERT_DEADLOOP_ENABLED    0x20
```
There's no definition for 6 to 7 bits. If we uses the 0xef, We might face the problem when new bit is added with some feature.

To prevent this, Let's spefiy what we need, NOT use for future usage :)

mgamal: Fair enough.

samimujawar: > (Note: The value is 0xff is also used in Juno, we might need to modify it there too).

@mgamal can you send a PR to fix this for Juno, please?
Also can you test this PR with RustSPMC on FVP and report back the result, please? You may also want to introduce an ASSERT (0) in the StMM code to test the behaviour (which should cause a CPU dead loop).

@LeviYeoReum Can you update the commit message to say that with the new settings a CpuDeadLoop() is expected when an ASSERT() fails, please?

LeviYeoReum: Hi @samimujawar,

> Can you update the commit message to say that with the new settings a CpuDeadLoop() is expected when an ASSERT() fails, please?

Yeap. I've updated.

> Note: The value is 0xff is also used in Juno, we might need to modify it there too

@mgamal 

What do you mean? ArmJuno.pkg uses PcdDebugPropertyMask using 0x2X (see the ArmVExpress.dsc.inc)

If you mean the StandaloneMm in the Juno, It's not yet upstream?

mgamal: Yes, you're right. StandaloneMM in Juno is not yet upstream. Please ignore my comment about Juno.

mgamal: @samimujawar Can confirm that with rust-spmc the execution continues after an ASSERT() without the patch and that execution stops after applying the fix.

LeviYeoReum:
@samimujawar I think it's ready to be merged.
Could you merge this please?

PR #810, opened 2025-07-01
Remove Overrides for AMD Server Platforms

pbgrimes:
Removing edk2 override files from code for AMD server platforms.  Going forward, any required overrides to tianocore/edk2 will be implemented in AMD's NDA AGESA code and will be considered temporary.  Temporary overrides will be removed from AMD's NDA AGESA code once changes that eliminate the need for an override are upstreamed to tianocore/edk2.

Temporary overrides for Turin can be found in the 1.0.0.7 Turin AGESA PI Package.

Note: A similar change and policy will be implemented for AMD embedded platforms.


abdattar (approved):
Looks good.

ndhillonm (approved):

PR #814, opened 2025-07-02
Platform/Ampere/buildfw.sh: add verbosity options

hrw:
The "build" command can provide user with more information during build. It can also be more quiet during operation.
       
This change adds "--silent" option to limit amount of output from build process.
    
It also adds "--verbose" option to give user more details during build.

### DIFF
--- a/Platform/Ampere/buildfw.sh
+++ b/Platform/Ampere/buildfw.sh
@@ -40,6 +40,8 @@ usage () {
   echo "  -l <kern>, --linuxboot <kern>    Build LinuxBoot firmware instead of full EDK2 with UEFI Shell, specifying path to flashkernel"
   echo "  -f, --flash                      Copy firmware to BMC and flash firmware (keeping EFI variables and NVPARAMs) after building"
   echo "  -F, --full-flash                 Copy firmware to BMC and flash full EEPROM (resetting EFI variables and NVPARAMs) after building"
+  echo "  -s, --silent                     Print less messages."
+  echo "  -v, --verbose                    Print more messages."
   echo ""
   echo "  Note: flash options require bmc.sh file with env vars BMC_HOST, BMC_USER and BMC_PASS defined"
   echo ""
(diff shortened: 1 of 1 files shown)


bexcran:
I'd prefer a verbosity parameter. 0 means silent; 1 means normal, 2 means extra verbose. It's what I've done for the AC04 build script that's not public yet.

e.g. parts from the script:
```
VERBOSE=1 # can be 0, 1 or 2
```
```
echo "  -v --verbose [0,*1*,2]                              Verbosity level. 0=none, 1=normal, 2=full"
```
```
  if [ "${VERBOSE}" = "0" ]; then
    BUILD_FLAGS="-s -q"
    MAKE_FLAGS="--quiet"
  elif [ "${VERBOSE}" = "1" ]; then
    BUILD_FLAGS=""
  else
    BUILD_FLAGS="--verbose"
  fi
```
```
    -v | --verbose)
      VERBOSE=$2
      shift 2
      ;;
```

MAKE_FLAGS is used when building BaseTools.

hrw:
> I'd prefer a verbosity parameter. 0 means silent; 1 means normal, 2 means extra verbose. It's what I've done for the AC04 build script that's not public yet.

I propose:

- -v 0 == --silent
- -v 1 == no options
- -v 2 == --verbose

This way those used to `build` script (most of EDK2 users(?)) have a way to make it silent (as they add `--silent/-s` as before) and those who want more info can add `--verbose/-v`

Instead of learning about levels of verbosity.

bexcran:
> > I'd prefer a verbosity parameter. 0 means silent; 1 means normal, 2 means extra verbose. It's what I've done for the AC04 build script that's not public yet.
> 
> I propose:
> 
>     * -v 0 == --silent
> 
>     * -v 1 == no options
> 
>     * -v 2 == --verbose
> 
> 
> This way those used to `build` script (most of EDK2 users(?)) have a way to make it silent (as they add `--silent/-s` as before) and those who want more info can add `--verbose/-v`
> 
> Instead of learning about levels of verbosity.

Sounds good. Could you update MAKE_FLAGS when -s is specified please?

hrw:
Can we merge it?

bexcran (approved):

bexcran:
@hrw  Sorry for the delay! I've just merged it.

PR #816, opened 2025-07-07
Various packages: add missing UefiCpuPkg dependency for ArmMmuLib

leiflindholm:
Commit 7516ab83ff67 ("Update ArmMmuLib library path") did what it says, but some packages in the tree have their own modules making use of ArmMmuLib, without previously having declared a dependency on UefiCpuPkg.

Add the missing dependencies.


leiflindholm:
@AjanZhong FYI. (It was not an obvious oversight.)

AjanZhong:
> FYI. (It was not an obvious oversight.)

ops, my fault. Missed these changes in PR.

PR #818, opened 2025-07-08
Update SiFive U540 Platform to use new FdtLib

JasonBrave:


JasonBrave:
@JohnAZoidberg @vlsunil @leiflindholm @changab

leiflindholm:
There are other users of OpensbiPlatformLib in the tree that should be updated at the same time.

RiscVPlatformPkg.dsc can't be built at the moment due to something being outdated w.r.t. opensbi.
@vlsunil can you look into that bit?

And the Freedom U500 fails on a missing TimerLib mapping. @JohnAZoidberg ?

But could you @JasonBrave update the FdtLib mappings in those .dscs anyway as part of this? (For whatever reason there are two entries in the U500 one.)

JasonBrave:
> And the Freedom U500 fails on a missing TimerLib mapping

Fixed

> But could you @JasonBrave update the FdtLib mappings in those .dscs anyway as part of this? (For whatever reason there are two entries in the U500 one.)

Updated FdtLib mapping for both SiFive platforms

JasonBrave:
One remaining issue is that RISCVOpensbiLib still depends on old FdtLib
image

leiflindholm:
> One remaining issue is that RISCVOpensbiLib still depends on old FdtLib 

Yeah.
That's a trickier one to resolve.

Someone told me the submodule-based opensbi was not the way forward and it was being replaced, but I've not been keeping track of the RISC-V port that much. Let's see if any of the other tagged people can chime in.

Anyway, the state of the platforms is better with these patches than without, so I can merge them tomorrow if no one speaks up.

JasonBrave:
Signed-off-by added

leiflindholm (approved):

vlsunil:
> There are other users of OpensbiPlatformLib in the tree that should be updated at the same time.
> 
> RiscVPlatformPkg.dsc can't be built at the moment due to something being outdated w.r.t. opensbi. @vlsunil can you look into that bit?
> 
Okay. Let me take a look. But I plan to remove this model of using opensbi as submodule completely.

leiflindholm:
> Okay. Let me take a look. But I plan to remove this model of using opensbi as submodule completely.

That is an entirely acceptable solution :)

vlsunil (approved):

PR #825, opened 2025-07-10
AMD/Xilinx: Fixes for ZynqMp stack size and SDHCI driver

mariam-elshakfy:
Following the comments on: https://github.com/tianocore/edk2/pull/10681
MmcDxe is not well-supported, so replace it with the suggested alternative:

> Ultimately, this should be replaced by Bus/Pci/SdMmcPciHcDxe/ used with NonDiscoverablePciDeviceDxe.

Also, increase the stack size to solve various runtime issues.

Tested on KV260.


michalsimek (approved):
All changes make sense and I have tested it on kv260.

abdattar (approved):
code changes good.

PR #826, opened 2025-07-14
Platform/ARM/VExpressPkg: change PcdMmBufferBase and SystemMemorySize

LeviYeoReum:
As TF-A is changed to reserved the "Normal shared area with Secure" in ARM_AP_TZC_DRAM1_BASE - 1MB  as much as 1MB, SystemMemory should be reduced 1MB and the PcdMmBufferBase is changed to be located in the area [0]. So change PcdMmBufferBase and SystemMemorySize.

Link: https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/40336 [0]

### DIFF
--- a/Platform/ARM/SgiPkg/SgiMemoryMap.dsc.inc
+++ b/Platform/ARM/SgiPkg/SgiMemoryMap.dsc.inc
@@ -73,10 +73,12 @@
   gArmSgiTokenSpaceGuid.PcdGpioController0Size|0x00010000
   gArmSgiTokenSpaceGuid.PcdGpioController0Interrupt|136
 
-  # System Memory (1GB - 16MB of Trusted DRAM at the top of the
-  # 32bit address space)
+  # System Memory (2GB - 16MB of Trusted DRAM at the top of the
+  # 32bit address space) and Normal shared area with
+  # Secure world is reserved 1MB from (2GB - 16MB).
+  # Therefore, set the default System Memory size to (2GB - 17MB)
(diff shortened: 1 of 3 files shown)


Platform/ARM/VExpressPkg/ArmVExpress-FVP-AArch64.dsc
```diff
@@ -140,9 +140,11 @@
   # System Memory
   # When RME is supported by the FVP the top 64MB of DRAM1 (i.e. at the top
   # of the 32bit address space) is reserved for four-world support in TF-A.
-  # Therefore, set the default System Memory size to (2GB - 64MB).
+  # And Normal shared area with Secure world is reserved 1MB from
```
samimujawar, line 145: Can you add a memory map diagram explaining the layout as a comment, please?

LeviYeoReum: Thanks. I'll add a diagram.

samimujawar (approved):

LeviYeoReum:
> Thank you for this patch. This change looks good to me. The TF-A patch series also seem to make similar changes to the TC and Neoverse-RD platforms. Is a similar update for those platforms needed as well? If so, can you point me to the PR, please?

In case of TC platform, It doesn't affect (since it define its own).
I'll add patch for Sgi platform.

Thanks

samimujawar:
@LeviYeoReum please send a separate PR for the remaining platforms.

PR #827, opened 2025-07-14
NetworkFeaturePkg: fix build failure

niruiyu:
When NETWORK_ENABLE is TRUE and NETWORK_ALLOW_HTTP_CONNECTIONS is TRUE,
 NetworkPkg/NetworkFixedPcds.dsc.inc specifies
 gEfiNetworkPkgTokenSpaceGuid.PcdAllowHttpConnections as TRUE.

But the PCD is set under [PcdsFixedAtBuild] section. The build.py will complain the following error:
Pcd (gEfiNetworkPkgTokenSpaceGuid.PcdAllowHttpConnections) defined in DSC is not declared in DEC files referenced
in INF files in FDF. Arch: ['IA32']

The commit fixes the failure by only setting the PCD in the [PcdsFixedAtBuild.$(DXE_ARCH)] section.

### DIFF
--- a/Features/Intel/Network/NetworkFeaturePkg/Include/NetworkFeature.dsc
+++ b/Features/Intel/Network/NetworkFeaturePkg/Include/NetworkFeature.dsc
@@ -32,10 +32,10 @@
 # PCD Section - list of EDK II PCD Entries modified by the feature.
 #
 ################################################################################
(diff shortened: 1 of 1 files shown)


niruiyu:
@lgao4 @nate-desimone @SaiChaganty any comments on the change?

PR #829, opened 2025-07-17
Fix build failures for BOSC and SOPHGO platforms

vlsunil:
BOSC and SOPHGO platforms don't build properly for below reasons.

1) Changes due to libfdt, CpuExceptionHandlerLib and RiscVFpuLib
2) Changes in PI spec and migrating from FIRMWARE_CONTEXT structure to a HOB due to this in edk2.

3) Unfortunately, these platforms are using drivers/libraries from other platforms (sometimes Ovmf as well). I am going to ask them to look at removing these dependencies. But since HiSilicon platforms are obsoleted and removed now, BOSC platforms which were using a library from HiSilicon no longer build. I have removed this dependency just to make sure platforms build without issue but BOSC maintainers need to make changes to find alternative to these.

The changes are only build tested. This is the first set of changes before I remove several opensbi related libraries/drivers.


vlsunil:
I reached out to SG2042 and BOSC platform owners and they have agreed to test and fix things after this PR. @leiflindholm : Please help with review and merge. Thanks!

vlsunil:
I felt there is no point just fixing the build issue. I will let platform owners to fix it properly with testing taking recently updated sifive platform as reference.

PR #836, opened 2025-07-29
Readme.md: Update obsolete build guidelines

b49020:
While working on a new platform port, these obsolete build guidelines were found during development and review. So lets follow the updated build guidelines for the main Readme too.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -62,7 +62,7 @@ W: https://github.com/tianocore/tianocore.github.io/wiki/Security
 # How to build (Linux Environment)
 
 ## Prerequisites
-The build tools themselves depend on Python (2) and libuuid. Most Linux systems
+The build tools themselves depend on Python (3) and libuuid. Most Linux systems
 will come with a Python environment installed by default, but you usually need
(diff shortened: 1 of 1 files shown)


Readme.md
```diff
@@ -171,21 +167,11 @@ After a successful build, the resulting images can be found in
...
-8
-```
-OK, so we have 8 CPUs - let's tell the build to use a little more than that:
-```
-$ NUM_CPUS=$((`getconf _NPROCESSORS_ONLN` + 2))
-```
-For the toolchain tag, use GCC5 for gcc version 5 or later, GCC4x for
-earlier versions, or CLANG35/CLANG38 as appropriate when building with clang.
+The main build process runs in parallel natively. For the toolchain tag, use
+GCC for gcc version 5 or later, GCC4x for earlier versions, or CLANG35/CLANG38
```
leiflindholm: Actually, the CLANG toolchain profiles are outdated too.
There's only CLANGPDB and CLANGDWARF now.

Sorry, but many thanks for cleaning up!

b49020: Fixed, thanks

leiflindholm (approved):

PR #847, opened 2025-08-12
Silicon/Ampere: Enable PCIe resizable BAR support

bexcran:
Modern graphics cards benefit from being able to negotiate a larger BAR size than they advertise by default. In testing, the Linux amdgpu driver crashed without it enabled. Since I'm not aware of any problems with it, enable it for all Ampere Altra platforms.

### DIFF
--- a/Silicon/Ampere/AmpereAltraPkg/AmpereAltraPkg.dsc.inc
+++ b/Silicon/Ampere/AmpereAltraPkg/AmpereAltraPkg.dsc.inc
@@ -463,6 +463,8 @@
   gArmPlatformTokenSpaceGuid.PL011UartClkInHz|1843200
   gArmPlatformTokenSpaceGuid.PL011UartInterrupt|0x62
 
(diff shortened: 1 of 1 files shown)


nhivp:
Thanks @bexcran for the great finding.
Hi anh @chuongtranle, could you comment on this enablement? I'm considering about PCIe MMIO32/MMIO64 allocation.

_But honestly, the amdgpu driver needs to be improved in case the resizable bar capability is not supported by firmware._

chuongtranle:
Resizeable BAR is an endpoint feature and no special handling by the RC is required. In Ampere Altra ARM64, endpoint can access entire MMIO64 space (size ~ 4TB per RC). This enable is just to make sure when firmware reads endpoint's configuration space and finds that endpoint has Resizeable BAR support, it is allowed the endpoint to have larger BAR size allocation.

chuongtranle (approved):
I am OK with this change.

Reviewed-by: Chuong Tran <chuong@os.amperecomputing.com>

PR #851, opened 2025-08-14
Platform/Qualcomm/RB3Gen2: Enable booting edk2 at EL2

b49020:
Many Qualcomm targets use Gunyah as a firmware-launched hypervisor, making EL2 unavailable for Non-secure firmware and operating systems, preventing use of alternative hypervisors.

Quoting Arm Base Boot Requirements here:

```
6.3.1.1. UEFI boot at EL2
  Systems must boot UEFI at EL2, to enable the installation of a hypervisor
  or a virtualization-aware operating system.
```

So lets use a TZ SMC call supported on these Qcom targets to clean up EL2/Gunyah configuration and transition to booting edk2 (UEFI) at EL2 exiting Gunyah. All the SMMU configuration gets cleaned up and handed over in EL2 in bypass mode. The edk2 is able to configure EL2 MMU/SMMU as you would expect on an Arm silicon executing edk2 in EL2.

Apart from that there are some build fixes included to follow library paths in upstream edk2.

### DIFF
--- a/Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.S
+++ b/Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.S
@@ -6,8 +6,34 @@
 
 #include <AsmMacroLib.h>
 #include <Library/ArmLib.h>
+#include "KodiakHelper.h"
 
 ASM_FUNC(ArmPlatformPeiBootAction)
-  ret
+// Many Qualcomm targets use Gunyah as a firmware-launched hypervisor, making
+// EL2 unavailable for Non-secure firmware and operating systems, preventing
+// use of alternative hypervisors.
+
+// Quoting Arm Base Boot Requirements here:
+//
+// 6.3.1.1. UEFI boot at EL2
+//   Systems must boot UEFI at EL2, to enable the installation of a hypervisor
+//   or a virtualization-aware operating system.
+
+// So lets use a TZ SMC call supported on these Qcom targets to clean up
+// EL2/Gunyah configuration and transition to booting edk2 (UEFI) at EL2
+// exiting Gunyah. All the SMMU configuration gets cleaned up and handed
+// over in EL2 in bypass mode. The edk2 is able to configure EL2 MMU/SMMU as
+// you would expect on an Arm silicon executing edk2 in EL2.
+
+  EL1_OR_EL2(x0)
+1:ldr   w0, =TZ_EL2_SWITCH_SMC_ID
+  ldr   w1, =TZ_EL2_SWITCH_PARAM_ID
+  mov   w2, wzr
+  mov   w3, wzr
+  ldr   w4, =TZ_EL2_SWITCH_PARAM2_EXIT_GUNYAH
+  smc   #0
+// From this point onwards executing in EL2
+
+2:ret
 
 ASM_FUNCTION_REMOVE_IF_UNREFERENCED
--- a/Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.h
+++ b/Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.h
@@ -1,0 +1,15 @@
+/** @file
+  Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef KODIAK_HELPER_H_
+#define KODIAK_HELPER_H_
+
+#define TZ_EL2_SWITCH_SMC_ID                0x02000121
+#define TZ_EL2_SWITCH_PARAM_ID              0x00000023
+#define TZ_EL2_SWITCH_PARAM2_EXIT_GUNYAH    0x1
+
+#endif /* KODIAK_HELPER_H_ */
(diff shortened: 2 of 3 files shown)


Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.S
```diff
@@ -7,7 +7,33 @@
 #include <AsmMacroLib.h>
 #include <Library/ArmLib.h>
 
+#define TZ_EL2_SWITCH_SMC_ID		0x02000121
```
leiflindholm: Could we move these out into a common header somewhere?

Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.S
```diff
@@ -7,7 +7,33 @@
...
+#define TZ_EL2_SWITCH_PARAM_ID		0x00000023
+#define TZ_EL2_SWITCH_PARAM2_KVM	0x1
+
 ASM_FUNC(ArmPlatformPeiBootAction)
-  ret
+// On Qcom targets supporting Gunyah, the edk2 (UEFI) is commonly loaded at
```
leiflindholm: I would prefer rewording this, since there is no fundamental reason for this, it's just part of the firware architecture on these platforms.

```
// Many Qualcomm targets use Gunyah as a firmware-launched hypervisor, making
// EL2 unavailable for Non-secure firmware and operating systems, preventing
// use of alternative hypervisors.
```

Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.S
```diff
@@ -7,7 +7,33 @@
 #include <AsmMacroLib.h>
 #include <Library/ArmLib.h>
 
+#define TZ_EL2_SWITCH_SMC_ID		0x02000121
+#define TZ_EL2_SWITCH_PARAM_ID		0x00000023
+#define TZ_EL2_SWITCH_PARAM2_KVM	0x1
```
leiflindholm: I know we have a tendency to conflate "don't use a firmware-owned hypervisor" with that weird little corner case of "use KVM", but that isn't actually what this is about.
We should give this a name that better describes the function.

Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.S
```diff
@@ -7,7 +7,33 @@
...
+// hypervisor stack to run.
+
+// Quoting Arm Base Boot Requirements here:
+//
+// 6.3.1.1. UEFI boot at EL2
+//   Systems must boot UEFI at EL2, to enable the installation of a hypervisor
+//   or a virtualization-aware operating system.
+
+// So lets use a TZ SMC call supported on these Qcom targets to clean up
+// EL2/Gunyah configuration and transition to booting edk2 (UEFI) at EL2
```
leiflindholm, line 23: I think this needs to document what actually happens, the current text is a bit handwavy. What happens to Gunyah?

Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.S
b49020: I am unsure if there will be much of reuse for these custom SMC calls with a separate header. Since in other places we will just be reusing standard `ArmMonitorCall()`.

leiflindholm: OK, if it's really not going to be used elsewhere, maybe it doesn't need to be a common header.

But the general rule in tianocore is #defines belong in header files, so could you break them out into a private .h?

Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.S
```diff
@@ -7,7 +7,33 @@
...
+// as the first step.
+
+  EL1_OR_EL2(x0)
+1:ldr   w0, =TZ_EL2_SWITCH_SMC_ID
+  ldr   w1, =TZ_EL2_SWITCH_PARAM_ID
+  mov   w2, wzr
+  mov   w3, wzr
+  ldr   w4, =TZ_EL2_SWITCH_PARAM2_KVM
+  smc   #0
+
```
leiflindholm, line 36: I think it's worth adding an explicit comment here that we will now be executing at EL2.

Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.S
b49020: Sure, I will follow that rule.

Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.S
b49020: Added further explanation.

Silicon/Qualcomm/KodiakPkg/Library/KodiakLib/KodiakHelper.h
```diff
@@ -0,0 +1,15 @@
...
+  Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+
+**/
+
+#ifndef KODIAK_HELPER_H_
+#define KODIAK_HELPER_H_
+
+#define TZ_EL2_SWITCH_SMC_ID			0x02000121
```
leiflindholm: I think once we bring in uncrustify to edk2-platforms CI, it will complain about the indentation difference. Could you align the columns?

b49020: Fixed

Issue #849, opened 2025-08-12
[Bug]: gFdtVariableGuid not declared in any DEC file


mdkinney:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

Platform/ARM/JunoPkg, Platform/Marvell/OdysseyPkg

### Which targets are impacted by this bug?

DEBUG, NO-TARGET, NOOPT, RELEASE

### Current Behavior

Platform\ARM\Drivers\FdtPlatformDxe\FdtPlatformDxe.inf uses gFdtVariableGuid, and that GUID has been removed from edk2 repo in PR https://github.com/tianocore/edk2/pull/11367. With this GUID not being defined in any DEC file, and platforms that uses FdtPlatformDxe.inf will fail to build.  This impacts the JunoPkg and OdysseyPkg that use FdtPlatformDxe.inf and could potentially impact other downstream platforms that use FdtPlatformDxe.inf.

### Expected Behavior

All platforms in edk2-platforms should build.

### Steps To Reproduce

Build JunoPkg or OdysseyPkg.  Build would break for gFdtVariableGuid not being defined.

### Build Environment

```markdown
All supported OS/Toolchains are impacted
```

### Version Information

```text
https://github.com/tianocore/edk2-platforms/commit/d1b297b2aa693fc723ac883866c925e9d04260c9
```

### Urgency

High

### Are you going to fix this?

Someone else needs to fix it

### Do you need maintainer feedback?

Maintainer feedback requested

### Anything else?

Recommend adding gFdtVariableGuid to Platform/ARM/Drivers/FdtPlatformDxe/FdtPlatformDxe.dec.

Alternatively, of JunoPkg and OdysseyPkg are no long required/supported, then those platforms could be removed along with the Platform\ARM\Drivers\FdtPlatformDxe component.

PR #852, opened 2025-08-14
Platform/ARM/Drivers: define gFdtVariableGuid in FdtPlatformDxe.dec

LeviYeoReum:
edk2 commit db5b28ed9e27 ("MdePkg, EmbeddedPkg: Move Fdt Table Guid to MdePkg")
removes gFdtVariableGuid while it moves the gFdtTableGuid to MdePkg.dec.

But, gFdtVariableGuid is used for development purpose to override
fdt by saving its path to load into the variable with gFdtVariableGuid.

gFdtVariableGuid is used by FdtPlatformDxe driver only and
no other place. So move gFdtVariableGuid into  FdtPlatformDxe.dec

This patch fixes build failure of ArmJuno Platform.


vbpandya:
This fixes #849

vbpandya (approved):
> gFdtVariableGuid is only used by FdtPlatformDxe driver and
> any other place. So move gFdtVariableGuid into  FdtPlatformDxe.dec

may be change this to 

gFdtVariableGuid is only used by FdtPlatformDxe driver and **not at**
any other place. So move gFdtVariableGuid into  FdtPlatformDxe.dec

mgamal:
I assume you'r referring to commit db5b28ed9e27 ("MdePkg, EmbeddedPkg: Move Fdt Table Guid to MdePkg"). Please correct this in the commit message.
Otherwise, looks good to me

Platform/ARM/Drivers/FdtPlatformDxe/FdtPlatform.h
```diff
@@ -26,7 +26,8 @@
 
 #include <Guid/Fdt.h>
 
-extern EFI_HANDLE mFdtPlatformDxeHiiHandle;
+extern EFI_GUID    gFdtVariableGuid;
```
samimujawar: Is this change needed?

LeviYeoReum: Yes. Otherwise C couldn't  get GUID with "Implict use"

LeviYeoReum:
> Thank you for this fix.
> 
> Not sure if the extern definition of gFdtVariableGuid is needed in the header file. Can you check and let me know? Otherwise this patch looks good to me.

Yes That's need. unless we declare. some copiler complains with"implicit usage".

Platform/ARM/Drivers/FdtPlatformDxe/FdtPlatform.h
samimujawar: Can you elaborate which compiler exhibits this issue, please? I have tried with GCC 14.3 rel1 and it seems to work.
To me it sounds like a tooling issue. Can you confirm if the declaration is not there in the Autogen.h file?

LeviYeoReum: You're right Since the AutoGen.h defines the as GUID type and EFI_GUID is defined as.

```
edk2/MdePkg/Include/Uefi/UefiBaseType.h <<EFI_GUID>>
             typedef GUID EFI_GUID;
```

So, there's no problem. i'll remove this line.

Thanks!

LeviYeoReum:
> > Thank you for this fix.
> > Not sure if the extern definition of gFdtVariableGuid is needed in the header file. Can you check and let me know? Otherwise this patch looks good to me.
> 
> Yes That's need. unless we declare. some copiler complains with"implicit usage".

@samimujawar. You're right. I've remove the extern declation.

samimujawar (approved):

Issue #855, opened 2025-08-19
[Bug]: ManageabilityPkg/Kcs: Remove duplicate files


cassalam:
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [x] Firmware
- [ ] Tool
- [ ] Unit Test

### What packages are impacted?

Features/ManageabilityPkg

### Which targets are impacted by this bug?

DEBUG, NO-TARGET, NOOPT, RELEASE

### Current Behavior

Files can be cloned from Features/ManageabilityPkg/Library/ManageabilityTransportKcsLib/Dxe to Features/ManageabilityPkg/Library/ManageabilityTransportKcsLib because this library can be used in both PEI and DXE.

The files under the /Dxe and /Common folders would be duplicates.

### Expected Behavior

No duplicate files in Features/ManageabilityPkg/Library/ManageabilityTransportKcsLib.

### Steps To Reproduce

N/A

### Build Environment

```markdown
- OS(s): Ubuntu 22.04, Windows 11
- Tool Chain(s): GCC12, VS2019
```

### Version Information

```text
Commit: bbc7f6b1a70e4e311a05274a2a475453caaf027f
```

### Urgency

Medium

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

Maintainer feedback requested

PR #857, opened 2025-08-20
ManageabilityPkg/Kcs: Remove duplicate files

cassalam:
Fixes https://github.com/tianocore/edk2-platforms/issues/855


changab:
> Just one comment, could you please add the elaboration in the commit message about the reason we keep the INF file under DXE? Thanks.

My bad. You already have that in the commit message.

pbgrimes (approved):

PR #862, opened 2025-08-29
Ampere: Update 'ATF' to 'TF-A'

bexcran:
Since it's been almost 10 years since the Trusted Firmware project rebranded from ATF to TF-A, go through and change all references in Ampere code from 'ATF' to 'TF-A' or 'TFA' depending on the context.


dkalowsk (approved):
about time

hrw (approved):
Looks Good To Me

builds fine, do not see missed entries

PR #863, opened 2025-08-31
Use DEBUG for FdtDxe

yangrongwei:
Print() will ASSERT in debug build, use DEBUG instead

### DIFF
--- a/Platform/RaspberryPi/Drivers/FdtDxe/FdtDxe.c
+++ b/Platform/RaspberryPi/Drivers/FdtDxe/FdtDxe.c
@@ -15,7 +15,6 @@
 #include <Library/FdtLib.h>
 #include <Library/MemoryAllocationLib.h>
 #include <Library/UefiBootServicesTableLib.h>
-#include <Library/UefiLib.h>
 #include <Protocol/RpiFirmware.h>
 #include <Guid/Fdt.h>
 #include <ConfigVars.h>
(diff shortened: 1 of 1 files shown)


ardbiesheuvel:
Why is this better?

benjamindoron:
It's asserting because `Print()` uses `gST->ConOut`, which is part of the ConPlatform, ConSplitter (and their children) stack. That's likely only available later. It'll actually 'assert' (as in, misbehave) on RELEASE builds too: it's a misconception that things 'only assert on the DEBUG build,' since ASSERTs are stripped in RELEASE.

Either way, it's proper for a driver to use `DEBUG`, not `Print`: that's for drivers and apps near the BDS phase to display things to the user, not debug information to the developer.

benjamindoron (approved):

leiflindholm:
Ugh, we need to get some CI in place in this repo. That commit message meets none of the requirements.

ardbiesheuvel:
> Ugh, we need to get some CI in place in this repo. That commit message meets none of the requirements.

Ah sorry, I missed that.

ardbiesheuvel:
This PR breaks the build so I am reverting it.

benjamindoron:
Yet another reason why we should have CI. I'll try debugging why when I can, if I remember, because that patch seemed good. Perhaps it was the removal of the UefiLib header.

ardbiesheuvel:
> Yet another reason why we should have CI.

+1

> I'll try debugging why when I can, if I remember, because that patch seemed good. Perhaps it was the removal of the UefiLib header.

No need to debug - it is just missing the `DEBUG_xxx` before the error string. But I'm kind of annoyed that people throw patches over the fence like this without even doing a test build.

benjamindoron:
> No need to debug - it is just missing the `DEBUG_xxx` before the error string. But I'm kind of annoyed that people throw patches over the fence like this without even doing a test build.

Wow. Now I'm embarrassed that I missed that too. Oh well: now we're using the right function, and it's no longer a mess.

PR #867, opened 2025-09-05
Support Altra platform secure boot

tamng-ampere:
Altra platform secure boot allows OEM implement and deploy their security certificates for platform secure boot and secure in-band firmware update. By enabling this feature, OEM will create a secure boot domain from SCP to UEFI.
For secure boot, firmware certificate will be created with DBB key. For secure in-band update, firmware signature will be created with DBU key.


chuongtranle:
@bexcran Need your feedback for ASRR and ADLINK platforms

PR #882, opened 2025-10-06
Platform/ASRockRack: Fix cache size value

bexcran:
Fix the cache size value in the SMBIOS Type 7 table. The first value is the 15-bit size, not the complete 16-bit value including the granularity.


leiflindholm:
@bexcran shouldn't you be in CODEOWNERS for this platform?

leiflindholm (approved):

PR #883, opened 2025-10-06
Platform/ASRockRack: Improve board settings configuration

bexcran:
Improve the board settings configuration by using the same values which ASRock Rack uses for their AptioV firmware.


RossComputerGuy (approved):
Seems to work fine on my ALTRAD8UD-1L2T system

leiflindholm (approved):

PR #884, opened 2025-10-13
MinPlatformPkg: TestPointCheckLib: Add AMD SmBase Check

BritChesley:
Added check for AMD vs Intel processors to get proper SM Base value.


Platform/Intel/MinPlatformPkg/Test/Library/TestPointCheckLib/SmmCheckPaging.c
```diff
@@ -191,7 +195,12 @@ GetSmBaseOnCurrentProcessor (
   SMBASE_SHARED_BUFFER  *SmBaseBuffer;
 
   SmBaseBuffer = Buffer;
-  SmBaseBuffer->SmBase = AsmReadMsr64 (MSR_IA32_SMBASE);
+  if (StandardSignatureIsAuthenticAMD ()) {
```
ChaselChiu, line 198: this introduces additional library dependency, please update INF too to include the lib, otherwise, the  build will fail with unresolved symbol.

BritChesley: Yes, I missed this. Added CpuLib in INF file

PR #886, opened 2025-10-20
Platform/ARM/VExpressPkg: Update virtio devices in ACPI

vbpandya:
This PR does two things

1. Add virtio net device as per the specification in FVP referemce guide
2. Update the size of virtio blk device to 64KB as per reference guide

Section 3.4 Base - memory map, Table 3-5 Base Platform memory map.

https://developer.arm.com/documentation/100966/latest/

### DIFF
--- a/Platform/ARM/VExpressPkg/ConfigurationManager/ConfigurationManagerDxe/AslTables/Dsdt.asl
+++ b/Platform/ARM/VExpressPkg/ConfigurationManager/ConfigurationManagerDxe/AslTables/Dsdt.asl
@@ -52,7 +52,7 @@ DefinitionBlock("DsdtTable.aml", "DSDT", 2, "ARMLTD", "ARM-VEXP", 1) {
       Name (_UID, 0)
 
       Name (_CRS, ResourceTemplate() {
-        Memory32Fixed (ReadWrite, 0x1c130000, 0x1000)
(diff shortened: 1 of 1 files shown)


Platform/ARM/VExpressPkg/ConfigurationManager/ConfigurationManagerDxe/AslTables/Dsdt.asl
```diff
@@ -56,5 +56,16 @@ DefinitionBlock("DsdtTable.aml", "DSDT", 2, "ARMLTD", "ARM-VEXP", 1) {
...
       })
     }
+
+    // VIRTIO net device
+    Device (VR01) {
+      Name (_HID, "LNRO0005")
+      Name (_UID, 1)
+
+      Name (_CRS, ResourceTemplate() {
+        Memory32Fixed (ReadWrite, 0x1c150000, 0x1000)
```
samimujawar: The size of the memory region should be 64KB, see https://developer.arm.com/documentation/100964/1128/Base-Platform/Base-Platform-memory/Base-Platform-memory-map

vbpandya: Nice catch!
While looking at it I also realized virtio blk size is also wrong, updated that as well.

PR #893, opened 2025-10-28
Drivers/OptionRomPkg: Deprecate gEfiScsiPassThruProtocolGuid

GowthamM9974:
This PR addresses the usage of gEfiScsiPassThruProtocolGuid that has been removed from the UEFI spec 2.10A and 2.11 but are still present in the codebase. To align with the updated spec the legacy ScsiPassThruProtocol has been deprecated and the implementation has been updated to only use ExtScsiPassThruProtocol, which provides enhanced support for device enumeration, target/LUN discovery, and broader SCSI compatibility.

Note: This PR should be merged before edk2 PR - https://github.com/tianocore/edk2/pull/11665 to avoid build errors.

### DIFF
--- a/Drivers/OptionRomPkg/AtapiPassThruDxe/AtapiPassThru.c
+++ b/Drivers/OptionRomPkg/AtapiPassThruDxe/AtapiPassThru.c
@@ -311,60 +311,27 @@ Returns:
 --*/
 {
   EFI_STATUS                      Status;
-  EFI_SCSI_PASS_THRU_PROTOCOL     *ScsiPassThru;
   EFI_EXT_SCSI_PASS_THRU_PROTOCOL *ExtScsiPassThru;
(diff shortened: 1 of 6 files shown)


niruiyu (changes requested):
How about removing the two PCDs: PcdSupportScsiPassThru, PcdSupportExtScsiPassThru

GowthamM9974:
> How about removing the two PCDs: PcdSupportScsiPassThru, PcdSupportExtScsiPassThru

Hi @niruiyu
Since ExtScsiPassThru is the only supported interface now, removing PcdSupportScsiPassThru and PcdSupportExtScsiPassThru makes sense. I’ve removed them accordingly and pushed the changes.

Would you mind help reviewing and proceed with the merge.
Thanks in advance!

PR #894, opened 2025-10-28
Ext4Pkg: remove unused variable in collation code

kaniini:
Iso639Language is not used anywhere in the code, so GCC flags it as an unused variable.


heatd (approved):
This was a leftover for the collation code removal that happened a few weeks/months ago. LGTM. Thank you!

PR #912, opened 2025-11-30
Platform/ARM/Morello: Remove unused variable

mikebeaton:
Found by compiling in GCC5 or GCC with stronger warnings currently PR-d in https://github.com/tianocore/edk2/pull/11761


mikebeaton:
How does merge/push work on edk2-platforms? Seems like in my earlier PRs everything was just (eventually) manually merged?

samimujawar:
I noticed the push label hence applied it. I was not sure if this would work. It seems we need to manually press the merge button which I have done now.

PR #918, opened 2025-12-02
Console Variable fix for TestPointCheckLib

shkrc:
ConInDev, ConOutDev & ErrOutDev variables have list of device paths which have terminal console devices appended at the end. For only the user selected terminal type, the device path will have the necessary protocols installed at any point in time. TestPointCheckLib tries to discover these non existing terminal types and reports as an error. This patch fixes this by checking if the device path is a terminal type, and if so, then ignore the terminal node and only check for the existence of the parent path.

### DIFF
--- a/Platform/Intel/MinPlatformPkg/Test/Library/TestPointCheckLib/DxeCheckDevicePath.c
+++ b/Platform/Intel/MinPlatformPkg/Test/Library/TestPointCheckLib/DxeCheckDevicePath.c
@@ -18,6 +18,8 @@ SPDX-License-Identifier: BSD-2-Clause-Patent
 #include <Library/MemoryAllocationLib.h>
 #include <Protocol/LoadedImage.h>
 #include <Protocol/DevicePath.h>
+#include <Guid/PcAnsi.h>
+#include <Guid/TtyTerm.h>
 
 VOID
 TestPointDumpDevicePath (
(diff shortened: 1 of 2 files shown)


shkrc:
@niruiyu requesting your reviews

Platform/Intel/MinPlatformPkg/Test/Library/TestPointCheckLib/DxeCheckDevicePath.c
```diff
@@ -71,6 +73,42 @@ TestPointDumpDevicePath (
...
+      (DevicePathNodeLength (Node) != sizeof (VENDOR_DEVICE_PATH)))
+  {
+    return FALSE;
+  }
+
+  for (Index = 0; Index < ARRAY_SIZE (mTerminalGuids); Index++) {
+    if (CompareGuid (&((VENDOR_DEVICE_PATH *)Node)->Guid, mTerminalGuids[Index])) {
+      return TRUE;
+    }
+  }
```
niruiyu, line 107: How about returning TRUE without checking the Guid?
Then the logic would support any future extension of new terminal types.

niruiyu (changes requested):
Can you check if the following logic works also?
I think it's simpler than the original logic.

```c
BOOLEAN
IsDevicePathExist (
  IN EFI_DEVICE_PATH_PROTOCOL          *DevicePath
  )
{
  EFI_DEVICE_PATH_PROTOCOL *DevicePathInstance;
  EFI_DEVICE_PATH_PROTOCOL *RemainingDevicePath;
  EFI_DEVICE_PATH_PROTOCOL *RemainingNode;
  EFI_HANDLE               Handle;
  BOOLEAN                  Exist;

  Exist = FALSE;
  RemainingDevicePath = DevicePath;
  do {
    DevicePathInstance = GetNextDevicePathInstance (&RemainingDevicePath, &Size);
    //
    // RemainingDevicePath points to the next instance.
    // DevicePathInstance points to a copy of the current instance.
    //
    if (DevicePathInstance == NULL) {
      break;
    }
    Exist         = FALSE;
    RemainingNode = DevicePathInstance;
    Status = gBS->LocateDevicePath (
                    &gEfiDevicePathProtocolGuid,
                    &RemainingNode,
                    &Handle
                    );
    if (!EFI_ERROR (Status) && (IsDevicePathEnd (RemainingNode) || IsTerminalVendorNode (RemainingNode))) {
      //
      // A handle is found whose device path
      //   either equals to the DevicePathInstance (when IsDevicePathEnd() is TRUE)
      //   or equals to the DevicePathInstance - TerminalNode (when IsTerminalVendorNode() is TRUE)
      //
      Exist = TRUE;
    }
    FreePool (DevicePathInstance);
  } while (Exist && RemainingDevicePath!= NULL);

  return Exist;
}
```

Platform/Intel/MinPlatformPkg/Test/Library/TestPointCheckLib/DxeCheckDevicePath.c
shkrc: Hi @niruiyu , i guess we need to check for the guids, because we are ignoring the child nodes only in the cases where the devices path is of a terminal device, for other cases, we dont need to ignore the child nodes. Hence, I think its better to define the terminal types

shkrc:
Hi with the above code suggestion, the issue I see only IsDevicePathNodeExist() does the validation whether the device actually exists. In your suggestion
` if (!EFI_ERROR (Status) && (IsDevicePathEnd (RemainingNode) || IsTerminalVendorNode (RemainingNode)
`
Lets assume there is a non-existing UART device that has a terminal child, then the above if condition would return true even when the device doesnt exist

shkrc:
Hi @niruiyu requesing comments

abdattar:
Gentle reminder, please provide the review comments.

abdattar:
Please review the patch and merge the changes if there are no further commits.

abdattar:
@ChaselChiu, @nate-desimone , @SaiChaganty, @lgao4 , @niruiyu , @ydong10 
Gentle reminder, please provide the review comments.

lgao4:
> Hi with the above code suggestion, the issue I see only IsDevicePathNodeExist() does the validation whether the device actually exists. In your suggestion `if (!EFI_ERROR (Status) && (IsDevicePathEnd (RemainingNode) || IsTerminalVendorNode (RemainingNode)` Lets assume there is a non-existing UART device that has a terminal child, then the above if condition would return true even when the device doesnt exist

I agree Ray solution. If the device doesn't exist, LocateDevicePath will return EFI_NOT_FOUND.

PR #924, opened 2025-12-23
Platform/ARM/SgiPkg: Enable PCIE

anandhkrishnau:
Enabled PCIE for FVP


vbpandya:
Can you please update the commit msg and PR description with more details ?
What is the background of this change ?
Is it fixing anything or if it is adding missing variables, Which models are you adding the value for ?

vbpandya:
Commit msg is also missing your sign off

PR #934, opened 2026-01-21
Vlv2 remove platform and silicon packages

mdkinney:
Draft PR Proposal to remove VLV2 from edk2-platforms.

Goal is to coordinate with other platforms and add a tag to edk2-platforms just before a collection of platform/silicon packages are removed, so there is a tag to access the set of archived platforms that are no longer supported.


bexcran:
Is there a newer Intel platform that people can buy if they want to work on firmware for a physical x86 device?
I wonder if there might still be value in keeping this around since it appears there are still a fairly large number available on the second-hand market.

mdkinney:
Will always be available in git history of edk2-platforms

mdkinney:
Close this draft PR.  

Changes were handled in more comprehensive #948

PR #938, opened 2026-02-03
MinPlatformPkg, Platforms: Relocate MinPlatformPkg for multi-arch support

vishalo:
Move MinPlatformPkg from Intel specific directory into directory Platform/MinPlatformPkg. This prepares the codebase for supporting additional architectures and for separating Intel specific logic from the common MinPlatformPkg framework.

Update build.cfg and build_bios.py to point to the new path for MinPlatformPkg under edk2-platforms/Platform.

Update CODEOWNERS and REVIWERES to new path.

Test:
Compiled AlderLake and QemuOpenBoarkdPkg.
Confirmed QemuOpenBoardPkg boots to shell.


nate-desimone (commented):
Hi @vishalo - `Features/MinPlatformPkg` isn't the right directory for MinPlatformPkg. MinPlatformPkg isn't a feature, it is architectural to the [EDK II Min-Platform Specification](https://tianocore-docs.github.io/edk2-MinimumPlatformSpecification/draft/). 

Appropriate places would be either `Platform/MinPlatformPkg` or putting MinPlatformPkg at the root of edk2-platforms. We should wait on something like this until after the Q1 stable tag is done because this package is has a very large number of downstream consumers.

vishalo:
Thank you for the review @nate-desimone - I agree that MinPlatformPkg's architectural nature makes Features/ not the ideal location.

I'm happy to move it to either Platform/MinPlatformPkg or edk2-platforms/MinPlatformPkg. My initial thinking (suggestion from Leif) was to separate it from Platform/Intel/ to signal multi-arch support, but I see your point.

Between the two options you mentioned, which would better align with the EDK II Min-Platform Specification's intent? 

Option 1: Platform/MinPlatformPkg - keeps it under the Platform hierarchy while removing the Intel-specific path

Option 2: Since we may have some silicon and architecture-specific code, what about:
  - edk2-platforms/MinPlatform/Silicon/
  - edk2-platforms/MinPlatform/Platform/
This would mirror the existing top-level structure and better support multi-arch in the future.

Fully agree with waiting until after the Q1 stable tag, given the number of downstream consumers.

Looking forward to hearing from others as well.

SaiChaganty:
Looking at the layout of edk2-platforms, it is a repository that allows companies to add their content (silicon, platform tools etc. ) unlike Edk2 repo which is more scoped around hosting cross community collaboration (barring maybe some exceptions :)).

I am leaning towards option 1 (unless we have details on the "silicon and architectural-specific code") primarily from a scope standpoint. MinPlatform Architecture deals with platform boot flow orchestration in a consistent manner.

I also think that we should have a good understanding of architecture specific Vs architecture agnostic content in this package before moving to edk2-platforms\Platform

leiflindholm:
@SaiChaganty I'm not going to get involved in where it goes, I mainly care about it becoming more widely used across architectures.

But I would like to put in a vote for _not_ figuring all aspects of which is architecture specific Vs architecture agnostic beforehand. That is something we can easily adjust after the fact.

Some things may want splitting up on an architecture level (which could use AARCH64/RISCV64/X64 subdirectories like in edk2), and some of it may want splitting up on a vendor level (Intel/AMD/Arm/Qualcomm).

And it may not be obvious which is which until there are more architectures and vendors making use of the code. And if we try to unravel all of that before we move anyway, the move may effectively be indefinitely stalled.

nate-desimone:
> Between the two options you mentioned, which would better align with the EDK II Min-Platform Specification's intent?

I also have a slight preference for `Platform/MinPlatformPkg`, but I'm not married to either option.

To me, the only things that look overtly Intel-specific in MinPlatformPkg is `MinPlatformPkg/FspWrapper/**` and `MinPlatformPkg/Tools/Fsp/**`. Those might be good candidates for a `Platform/Intel/IntelPlatformPkg` or something like that. The rest of it seems mostly generic to me. The ultimate purpose of the package is to be middleware that:

1. Adds consistency across platforms
2. Makes it easier use of EDK II.

I believe most of the code in MinPlatformPkg furthers that goal in a silicon agnostic manner.

vishalo:
Thank you @leiflindholm .

> I am leaning towards option 1 (unless we have details on the "silicon and architectural-specific code") primarily from a scope standpoint.
>I also have a slight preference for Platform/MinPlatformPkg, but I'm not married to either option.

Thank you @SaiChaganty  and @nate-desimone . I've updated patch to move to Platform/MinPlatformPkg.

vishalo:
Hi @nate-desimone,

I will rebase once https://github.com/tianocore/edk2-platforms/pull/948 merges.

Regarding the Intel-specific components (MinPlatformPkg/FspWrapper/** and MinPlatformPkg/Tools/Fsp/**): I'm inclined to handle this separation to Platform/Intel/IntelPlatformPkg as a follow-up PR to keep this change focused on the relocation. However, I'm happy to include it here if you prefer. What are your thoughts?

leiflindholm:
@vishalo did you accidentally close this PR and want it reopened?

vishalo:
> @vishalo did you accidentally close this PR and want it reopened?

Yes @leiflindholm . Did not mean to close it.

leiflindholm:
@ardbiesheuvel @makubacki @mdkinney could one of you with write access reopen this PR please?

mdkinney:
Looks like this PR is in a bad state.  There are no files changed.

leiflindholm:
@vishalo please push your latest set to your move_minplat branch.

vishalo:
> @vishalo please push your latest set to your move_minplat branch.

https://github.com/tianocore/edk2-platforms/commit/69b22ee94ea5a631e522a7324b8784911431020e

Platform/Intel/build_bios.py
```diff
@@ -121,7 +121,8 @@ def pre_build(build_config, build_type="DEBUG", silent=False, toolchain=None, sk
                                                config["WORKSPACE_FSP_BIN"])
 
     # add to package path
-    config["PACKAGES_PATH"] = config["WORKSPACE_PLATFORM"]
+    config["PACKAGES_PATH"] = os.path.join(config["WORKSPACE"], "edk2-platforms", "Platform")
```
leiflindholm: This feels unrelated?

leiflindholm (commented):
The diff is massive and not just showing moves - have you changed line endings too?

Platform/Intel/build_bios.py
vishalo: We have to update this to add edk2-platforms/Platform/MinPlatformPkg to PACKAGES_PATH.

vishalo:
> The diff is massive and not just showing moves - have you changed line endings too?

No, I did not update line endings.

leiflindholm:
> > The diff is massive and not just showing moves - have you changed line endings too?
> 
> No, I did not update line endings.

Then you need to have a serious talk to whoever is messing with your repository.

```
edk2-platforms$ file Platform/Intel/MinPlatformPkg/Acpi/AcpiSmm/AcpiMm.c
Platform/Intel/MinPlatformPkg/Acpi/AcpiSmm/AcpiMm.c: c program text, ASCII text, with CRLF line terminators
edk2-platforms$ git checkout --track vishalo/move_minplat
branch 'move_minplat' set up to track 'vishalo/move_minplat'.
Switched to a new branch 'move_minplat'
edk2-platforms$ file Platform/MinPlatformPkg/Acpi/AcpiSmm/AcpiMm.c
Platform/MinPlatformPkg/Acpi/AcpiSmm/AcpiMm.c: c program text, ASCII text
```

Platform/Intel/build_bios.py
leiflindholm: We do for how _we_ want to use MinPlatformPkg, but we don't for _moving_ MinPlatform, so it shouldn't be in this PR.

vishalo: Discussed offline with Leif - aligning to how other variables are setup.

vishalo:
> > > The diff is massive and not just showing moves - have you changed line endings too?
> > 
> > 
> > No, I did not update line endings.
> 
> Then you need to have a serious talk to whoever is messing with your repository.
> 
> ```
> edk2-platforms$ file Platform/Intel/MinPlatformPkg/Acpi/AcpiSmm/AcpiMm.c
> Platform/Intel/MinPlatformPkg/Acpi/AcpiSmm/AcpiMm.c: c program text, ASCII text, with CRLF line terminators
> edk2-platforms$ git checkout --track vishalo/move_minplat
> branch 'move_minplat' set up to track 'vishalo/move_minplat'.
> Switched to a new branch 'move_minplat'
> edk2-platforms$ file Platform/MinPlatformPkg/Acpi/AcpiSmm/AcpiMm.c
> Platform/MinPlatformPkg/Acpi/AcpiSmm/AcpiMm.c: c program text, ASCII text
> ```

Thank you @leiflindholm  !

Platform/Intel/build_bios.py
leiflindholm: (And agreed that with that update, this change _does_ belong in this PR.)

Platform/Intel/build_bios.py
```diff
@@ -408,7 +411,8 @@ def build(config):
             if re.search(pattern, item):
                 os.remove(os.path.join(file_dir, item))
 
-        command = [sys.executable, os.path.join(config['WORKSPACE_PLATFORM'],
+        command = [sys.executable, os.path.join(config['WORKSPACE'],
```
leiflindholm: This needs aligning with the above changes.

leiflindholm (changes requested):

vishalo:
Moved to using MIN_PACKAGE_TOOLS to get to Fsp scripts.

leiflindholm (approved):

vishalo:
Hi @nate-desimone, @SaiChaganty,

Thank you again for the helpful feedback on this PR. I've updated it to move MinPlatformPkg to `Platform/MinPlatformPkg` as we discussed - this is ready for review.

leiflindholm:
@nate-desimone @SaiChaganty @mdkinney if no further comments, can I go ahead and merge?

Platform/Intel/build_bios.py
nate-desimone, line 133: `MIN_PACKAGE_TOOLS` is consumed as a relative path but never converted to an absolute path, breaking the established pattern.

The new code passes `config['MIN_PACKAGE_TOOLS']` directly to `os.path.join()` in `build()` without ever resolving it to an absolute path. The old code was safe because it composed the command path from `config['WORKSPACE_PLATFORM']` which was already absolute. The replacement breaks that invariant.

Please add the following to `pre_build()`, alongside the other `WORKSPACE_*` conversions:

```python
config["MIN_PACKAGE_TOOLS"] = os.path.join(config["WORKSPACE"], config["MIN_PACKAGE_TOOLS"])
```

vishalo: Thank you @nate-desimone  ! I've updated it in latest commit.

nate-desimone (approved):
Hi @vishalo - the PR title and description still reference the "Features directory". Could you please correct that. Other than that, the code looks good! Approved!

vishalo:
> Hi @vishalo - the PR title and description still reference the "Features directory". Could you please correct that. Other than that, the code looks good! Approved!

I've updated this. Thank you @nate-desimone  !

abdattar (approved):
looks good

shkrc (approved):

PR #980, opened 2026-04-21
Platform/Qualcomm/RB3Gen2: Switch ELF gen script from GCC5 to GCC

vishalo:
GCC5 is deprecated in edk2. Update the RB3Gen2 ELF generation script to use GCC_AARCH64_PREFIX, matching upstream behavior.

Reference: https://github.com/tianocore/edk2/pull/12055


b49020 (approved):
Looks good to me

leiflindholm (approved):

PR #981, opened 2026-04-27
When configuring the system to clear the user password,

Adadxz:
the direct variable update in PasswordSmmInit fails with an EFI_NOT_AVAILABLE_YET error. This occurs because the initial NV variable write relies on the EFI_VARIABLE_WRITE_ARCH_PROTOCOL, which has not yet been installed during the early DXE/SMM init phase.


SaiChaganty:
This appears to be a partial solution to ensure there are no errors if EFI_VARIABLE_WRITE_ARCH_PROTOCOL is not installed. 
I think we also need to how to ensure EFI_VARIABLE_WRITE_ARCH_PROTOCOL is installed.

SaiChaganty:
Please rephrase the commit message to make the intent of this PR clear.

SaiChaganty (approved):

PR #1000, opened 2026-06-25
MinPlatformPkg/TestPointLib: Prevent calling FreePool on a NULL pointer

antklein:
Aip->GetSupportedTypes may return EFI_SUCCESS when 0 supported types are found. This results in an attempt to call FreePool on a NULL pointer. To fix this explicitly verify that InfoTypesBuffer is not NULL before calling FreePool on the pointer.

### DIFF
--- a/Platform/MinPlatformPkg/Test/Library/TestPointLib/DxeTestPoint.c
+++ b/Platform/MinPlatformPkg/Test/Library/TestPointLib/DxeTestPoint.c
@@ -72,11 +72,13 @@ InternalTestPointFindAip (
     //
     // Check AIP
     //
-    Status = Aip->GetSupportedTypes (
-                    Aip,
-                    &InfoTypesBuffer,
-                    &InfoTypesBufferCount
-                    );
+    InfoTypesBuffer      = NULL;
+    InfoTypesBufferCount = 0;
+    Status               = Aip->GetSupportedTypes (
+                                  Aip,
+                                  &InfoTypesBuffer,
+                                  &InfoTypesBufferCount
+                                  );
     if (EFI_ERROR (Status)) {
       continue;
     }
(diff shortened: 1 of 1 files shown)


antklein:
@nate-desimone, @SaiChaganty, @lgao4, @ydong10 Can I get your help reviewing this PR?

Platform/MinPlatformPkg/Test/Library/TestPointLib/DxeTestPoint.c
```diff
@@ -72,12 +72,20 @@ InternalTestPointFindAip (
...
+    Status               = Aip->GetSupportedTypes (
+                                  Aip,
+                                  &InfoTypesBuffer,
+                                  &InfoTypesBufferCount
+                                  );
+    if (EFI_ERROR (Status) ||
+        (InfoTypesBuffer == NULL) ||
+        (InfoTypesBufferCount == 0))
+    {
+      if (InfoTypesBuffer != NULL) {
```
SaiChaganty: Wouldn't it be simple to just check for InfoTypeBuffer for (!= NULL) at line 99 ?

antklein: Yes, that is an option as well. This change lets the loop exit slightly earlier. But I do like the simplicity of making a smaller change.

antklein: It also depends on how much trust you have in the provider of the `GetSupportedTypes` API. It's possible to return a NULL InfoTypesBuffer, with a count greater than 0. In this scenario, the for loop for `AipCandidate` will dereference a NULL pointer.

But in most scenarios, I do agree that the simple NULL check change before calling FreePool is sufficient.

SaiChaganty (approved):

antklein:
@SaiChaganty Can I get your help to merge this PR now that it has been approved?

PR #1001, opened 2026-06-25
Silicon/Qualcomm: Introduce QualcommSiliconPkg

vishalo:
This patch series introduces QualcommSiliconPkg, to support MinPlatform implementation for Qualcomm SOC.
Update maintainers and reviewers to match.

PR #1003, opened 2026-06-26
Platform/Qualcomm: Introduce QualcommPlatformPkg 

vishalo:
This patch series introduces QualcommPlatformPkg, to support MinPlatform implementation for Qualcomm SOC.


vishalo:
@aiyengar-qc

vishalo:
Addressed comments from https://github.com/tianocore/edk2-platforms/pull/1001 that are relevant to this PR too.

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
```diff
@@ -0,0 +1,153 @@
...
+    - QSEE  - Qualcomm Secure Execution Environment
+    - RPM   - Resource Power Manager
+    - SBL   - Secondary Boot Loader
+
+**/
+
+#pragma once
+
+/*
+ * Following magic number indicates the boot shared imem region is initialized
```
leiflindholm: We can induce what imem from the file header, but it still needs to be in the glossary, which needs to explain what it is and what it is for.

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
```diff
@@ -0,0 +1,153 @@
+/** @file
+
+  This header file contains declarations and definitions for Boot shared
+  internal memory cookies.
+
+  These cookies are shared between boot and other subsystems.
```
leiflindholm: What's a boot subsystem in this context?

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
```diff
@@ -0,0 +1,153 @@
...
+
+**/
+
+#pragma once
+
+/*
+ * Following magic number indicates the boot shared imem region is initialized
+ * and the content is valid
+ */
+#define BOOT_SHARED_IMEM_MAGIC_NUM  0xC1F8DB40
```
leiflindholm: This is an exported header - (all) symbols in it need to be suitably namespaced.

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
```diff
@@ -0,0 +1,153 @@
...
+   XBL that it needs to be backed up for delayed ramdump scenario */
+#define QSEE_DLOAD_DUMP_SHARED_IMEM_OFFSET     0x748
+#define QSEE_DLOAD_DUMP_SHARED_IMEM_MAGIC_NUM  0x44554d50
+
+/* Shared IMEM offset and magic number used by QHEE to indicate to
+   XBL that it needs to be backed up for delayed ramdump scenario.
+   Use magic number 0x44554d50 instead of legacy 0xC1F8DB42 to enable
+   dumping multiple TZ segments instead of legacy TZ binary blob
+*/
+#define QHEE_DLOAD_DUMP_SHARED_IMEM_MAGIC_NUM  0x44554d50
```
leiflindholm: What is QHEE? An while we're at it, what's a DLOAD?

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
```diff
@@ -0,0 +1,153 @@
...
+     and the content is valid.*/
+  UINT32    SharedImemMagic;
+
+  /* Number to indicate what version of this structure is being used */
+  UINT32    SharedImemVersion;
+
+  /* Pointer that points to etb ram dump buffer, should only be set by HLOS */
+  UINT64    EtbBufAddr;
+
+  /* Region where HLOS will write the l2 cache dump buffer start address */
```
leiflindholm: This sounds very much like some special case at a historic point in time that unfortunately got codified into a protocol that needs to maintain backwards compatibility forever. (I.e. what happens when there's a core with an L3, but even before that, in a system with dozens of L2 caches, which particular one are we talking about.)

Is this still in use, and if so what does it actually mean on an architectural level?

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
```diff
@@ -0,0 +1,153 @@
...
+  UINT64    L2CacheDumpBuffAddr;
+
+  /* When SBL which is A32 allocates the 64bit pointer above it will only
```
leiflindholm: What is A32? 
"Will" is passive voice. Passive voice is not documentation. Please rewrite in active voice.

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
```diff
@@ -0,0 +1,153 @@
...
+ * Following structure defines all the cookies that have been placed
+ * in boot's shared imem space.
+ * The size of this struct must NOT exceed SHARED_IMEM_BOOT_SIZE
+ */
+typedef struct {
+  /* Magic number which indicates boot shared imem has been initialized
+     and the content is valid.*/
+  UINT32    SharedImemMagic;
+
+  /* Number to indicate what version of this structure is being used */
```
leiflindholm: Can we get rid of all the "say exactly what the field name already says, but in prose" comments? (But keep the ones that add relevant context.)

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
```diff
@@ -0,0 +1,153 @@
...
+  UINT32    SharedImemVersion;
+
+  /* Pointer that points to etb ram dump buffer, should only be set by HLOS */
+  UINT64    EtbBufAddr;
+
+  /* Region where HLOS will write the l2 cache dump buffer start address */
+  UINT64    L2CacheDumpBuffAddr;
+
+  /* When SBL which is A32 allocates the 64bit pointer above it will only
+     consume 4 bytes.  When HLOS running in A64 mode access this it will over
```
leiflindholm: What?
The field is 64 bits in size. How would it overflow?

leiflindholm: While you're at the active voice rewrite, please get rid of the horrific use of "consume" which is not how any of this works.

Platform/Qualcomm/QualcommPlatformPkg/Include/Library/RamPartitionTableLib.h
```diff
@@ -0,0 +1,200 @@
...
+#include <RamPartition.h>
+
+/**
+  Read RAM partition table and cache entries.
+
+  @retval  EFI_SUCCESS  Successfully initialized the library.
+  @retval  Other        Initialization failed.
+
+**/
+EFI_STATUS
```
leiflindholm, line 29: These all seem to be missing EFIAPI (please sanity check through all of the PRs).
(applies throughout)

Platform/Qualcomm/QualcommPlatformPkg/Include/Library/RamPartitionTableLib.h
```diff
@@ -0,0 +1,200 @@
...
+InitRamPartitionTableLib (
```
leiflindholm: This is an exported library - symbols need to do some name space sanitisation. (applies throughout)

Platform/Qualcomm/QualcommPlatformPkg/Include/Library/RamPartitionTableLib.h
```diff
@@ -0,0 +1,200 @@
...
+  @param[in,out]  Version             Major version on output.
+
+  @retval  EFI_SUCCESS             Successfully retrieved version.
+  @retval  EFI_NOT_FOUND           RAM partition table is not available.
+  @retval  EFI_INVALID_PARAMETER   The RAM partition table is invalid.
+
+**/
+EFI_STATUS
+GetRamPartitionVersion (
+  IN VOID        *pRamPartitionTable,
```
leiflindholm: There is technically an exemption for the 'p' form of Hungarian notation from the overall ban: https://tianocore-docs.github.io/edk2-CCodingStandardsSpecification/release-2.40/4_naming_conventions/44_identifiers.html#4431-use-of-hungarian-notation-is-not-allowed (which we should probably get around to abolishing).

There is however no exception for using it on a single pointer in a library, ignoring all others.

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+
+**/
+EFI_STATUS
+GetRamPartitionVersion (
+  IN VOID        *pRamPartitionTable,
+  IN OUT UINT32  *Version
+  )
+{
+  /* v0 and v1 have same header info */
+  USABLE_RAM_PART_TABLE_TYPE  pRamPartTable;
```
leiflindholm: Don't just abbreviate the input parameter name for the local variable.
(And don't use Hungarian notation.)
This is a small function - PartitionTable, or even Table would be entirely unambiguous.

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+**/
+STATIC
+EFI_STATUS
+GetRamPartitionTable (
+  IN OUT VOID    **pRamPartitionTable,
+  IN OUT UINT32  *Version
+  )
+{
+  EFI_STATUS  Status;
+  UINT32      RamPartitionBuffSz;
```
leiflindholm: Size

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+      (pRamPartTable->Magic2 == RAM_PART_MAGIC2))
+  {
+    *Version = pRamPartTable->Version;
+    return EFI_SUCCESS;
+  } else {
+    return EFI_NOT_FOUND;
+  }
+}
+
+#define UNSUPPORTED_BELOW_VER  1
```
leiflindholm: Please move to top of file (or a private header).

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+  *pRamPartitionTable = SmemGetAddr (SmemUsableRamPartitionTable, (UINT32 *)&RamPartitionBuffSz);
+  if (*pRamPartitionTable == NULL) {
+    /* NOTE: We should be here only if SMEM is not initialized */
+    DEBUG ((DEBUG_ERROR, "WARNING: Unable to read memory partition table from SMEM\n"));
+    return EFI_NOT_READY;
+  }
+
+  Status = GetRamPartitionVersion (*pRamPartitionTable, Version);
+  if (Status != EFI_SUCCESS) {
+    /* Invalid RAM Partition, return early */
```
leiflindholm: Kind of redundant.

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+    return Status;
+  }
+
+  if (*Version < UNSUPPORTED_BELOW_VER) {
+    DEBUG ((DEBUG_WARN, "WARNING: Using deprecated RAM partition table !\n"));
```
leiflindholm: If we're going into CpuDeadLoop, we're not exactly "Using" it, are we?

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+
+  @param[out]  MemoryCapacity  Pointer to receive the total SDRAM memory size.
+
+  @retval  EFI_SUCCESS             Memory capacity retrieved successfully.
+  @retval  EFI_INVALID_PARAMETER   MemoryCapacity is NULL.
+  @retval  Other                   Error occurred during retrieval.
+
+**/
+EFI_STATUS
+GetInstalledSDRAMMemory (
```
leiflindholm: Sdram

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+  UINTN  *MemoryCapacity
+  )
+{
+  EFI_STATUS                  Status;
+  VOID                        *pRamPartitionTable;
+  USABLE_RAM_PART_TABLE_TYPE  pRamPartTable;
```
leiflindholm: I was about to repeat my comment from above about not just using an abbreviated form of one variable for another. But in this case there's literally no point to having both. Please delete one.

I see the antipattern repeated in other functions below, please address there too.

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+
+  if (MemoryCapacity == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  SetMem (EntryList, sizeof (EntryList), 0);
+  EntryCount = RAM_NUM_PART_ENTRIES;
+  Index      = 0;
+
+  Status = GetRamPartitions (&EntryCount, &EntryList[0]);
```
leiflindholm: `&EntryList[0]`?
You mean `EntryList`?

Please address throughout.

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+
+  if (FirstBank == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  if (mRamPartitionTableEntryCount > RAM_NUM_PART_ENTRIES) {
+    return EFI_DEVICE_ERROR;
+  }
+
+  Index      = 0;
```
leiflindholm: Done at for-loop entry.

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+  if ((*EntryCount > 0) && (EntryTable == NULL)) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  if (Count > *EntryCount) {
+    *EntryCount = Count;
+    return EFI_BUFFER_TOO_SMALL;
+  }
+
+  for ( Index = 0; Index < Count; Index++ ) {
```
leiflindholm: Stray space before Index.
Actually see several random variations of whitespace usage around for-loops in this file. Did you run uncrustify?

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+  EFI_STATUS                     Status;
+  UINT32                         Index;
+  USABLE_RAM_PART_TABLE_TYPE_V1  pRamPartTablev1;
+
+  if (pRamPartitionTable == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  Status          = EFI_SUCCESS;
+  Index           = 0;
```
leiflindholm: OK, this is a pattern now - please drop multi-index-initializations throughout.

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+
+      EndAddr = StartAddr + Length;
+
+      /* Handle boundary condition for UINT32 rollover */
+      if (EndAddr == 0x100000000ULL) {
+        Length  = Length - 0x200000;
+        EndAddr = StartAddr + Length;
+      }
+
+      if (mRamPartitionTableEntryCount >= RAM_NUM_PART_ENTRIES) {
```
leiflindholm: Why is this not done before the address calculations?

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+  for ( Index = 0; Index < pRamPartTablev1->NumPartitions; Index++ ) {
+    RAM_PARTITION_TYPE  PartitionType;
+    UINT64              StartAddr;
+    UINT64              Length;
+
+    PartitionType = pRamPartTablev1->RamPartEntryV1[Index].PartitionType;
+    StartAddr     = pRamPartTablev1->RamPartEntryV1[Index].StartAddress;
+    Length        = pRamPartTablev1->RamPartEntryV1[Index].Length;
+
+    if ((pRamPartTablev1->RamPartEntryV1[Index].PartitionType == RamPartitionSysMemory) &&
```
leiflindholm: Please break either branch of this if statement out into separate helpers.

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+  Index           = 0;
+  pRamPartTablev1 = (USABLE_RAM_PART_TABLE_TYPE_V1)pRamPartitionTable;
+
+  /* Parse the DDR partition table, and fill in table for only DDR entries */
+  for ( Index = 0; Index < pRamPartTablev1->NumPartitions; Index++ ) {
+    RAM_PARTITION_TYPE  PartitionType;
+    UINT64              StartAddr;
+    UINT64              Length;
+
+    PartitionType = pRamPartTablev1->RamPartEntryV1[Index].PartitionType;
```
leiflindholm: Could we tidy this function up a bit with temporary variables for 
`RamPartTablev1->RamPartEntryV1[Index]`,
`mRamPartitionTable[mRamPartitionTableEntryCount]`, and
`mPreloadRamPartitionTable[mPreloadRamPartitionTableEntryCount]`?

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+  VOID
+  )
+{
+  EFI_STATUS                  Status;
+  UINT32                      Index;
+  UINT32                      Version;
+  VOID                        *pRamPartitionTable;
+  USABLE_RAM_PART_TABLE_TYPE  pRamPartTable;
+
+  SetMem (mRamPartitionTable, sizeof (mRamPartitionTable), 0);
```
leiflindholm: ZeroMem?

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+  if (pRamPartTable == NULL) {
+    return EFI_INVALID_PARAMETER;
+  }
+
+  /* Parse the DDR partition table, and fill in table for only DDR entries */
+  for ( Index = 0; Index < pRamPartTable->NumPartitions; Index++ ) {
+    RAM_PARTITION_TYPE  PartitionType;
+    UINT64              StartAddr;
+
+    PartitionType = pRamPartTable->RamPartEntry[Index].PartitionType;
```
leiflindholm: Temporary variables for 
`RamPartTable->RamPartEntry[Index]`,
`mRamPartitionTable[mRamPartitionTableEntryCount]`, and 
`mPreloadRamPartitionTable[mPreloadRamPartitionTableEntryCount]`?

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+    StartAddr     = pRamPartTable->RamPartEntry[Index].StartAddress;
+
+    if ((pRamPartTable->RamPartEntry[Index].PartitionType == RamPartitionSysMemory) &&
```
leiflindholm: Split if and else bodies into helper functions?

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
```diff
@@ -0,0 +1,783 @@
...
+      Length = pRamPartTable->RamPartEntry[Index].AvailableLength;
+
+      EndAddr = StartAddr + Length;
+      if (FixedPcdGet64 (PcdMaxMemory) != 0) {
+        if (EndAddr > FixedPcdGet64 (PcdMaxMemory)) {
+          continue;
+        }
+      }
+
+      if (mRamPartitionTableEntryCount >= RAM_NUM_PART_ENTRIES) {
```
leiflindholm: Why do this check after above arithmetic?

Platform/Qualcomm/QualcommPlatformPkg/Library/DebugPecoffExtraActionTrace32Lib/DebugPecoffExtraActionTrace32Lib.c
```diff
@@ -0,0 +1,816 @@
...
+#include <Library/BaseLib.h>
+#include <Library/BaseMemoryLib.h>
+#include <Library/CacheMaintenanceLib.h>
+#include <Library/DebugLib.h>
+#include <Library/MemoryAllocationLib.h>
+#include <Library/PeCoffExtraActionLib.h>
+#include <Library/PeCoffLib.h>
+#include <Library/PrintLib.h>
+
+#define ADDRESS_FORMAT_STRING  "0x%016lx"
```
leiflindholm: Could you break all the macros and struct definitions out into a local header?

Platform/Qualcomm/QualcommPlatformPkg/Library/DebugPecoffExtraActionTrace32Lib/DebugPecoffExtraActionTrace32Lib.c
```diff
@@ -0,0 +1,816 @@
...
+  pointed to by SrcStr to their lowercase equivalents.
+
+  @param[in, out]  SrcStr  Null-terminated source string to convert.
+
+  @return  Pointer to the converted string (same as SrcStr).
+
+**/
+STATIC
+CHAR8 *
+AsciiToLower (
```
leiflindholm, line 325: This feels like a generic function that should live in MdePkg/Library/BaseLib/String.c, not implemented down in a leaf.
Please either implement and submit that, or change the code to use the existing AsciiCharToUpper () for case insensitivity handling.

Platform/Qualcomm/QualcommPlatformPkg/Library/DebugPecoffExtraActionTrace32Lib/DebugPecoffExtraActionTrace32Lib.c
```diff
@@ -0,0 +1,816 @@
...
+  @param[in]  ImageContext  Pointer to the image context structure.
+
+**/
+VOID
+ProcessImageUnload (
+  IN PE_COFF_LOADER_IMAGE_CONTEXT  *ImageContext
+  )
+{
+  //
+  // Append-only model: do not remove entries on unload.
```
leiflindholm, line 671: Is this a euphemism for TODO?

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+#include <Library/PcdLib.h>
+#include <Library/RamPartitionTableLib.h>
+#include <Library/SmemLib.h>
+
+#include <PiPei.h>
+#include <Pi/PiHob.h>
+#include <Pi/PiBootMode.h>
+
+#include <MemRegionInfo.h>
+#include <PlatformConfiguration.h>
```
leiflindholm: Highlight this is local by using ""?

Platform/Qualcomm/QualcommPlatformPkg/Include/MemRegionInfo.h
```diff
@@ -0,0 +1,82 @@
+/** @file
+  Memory Region Information Definitions.
```
leiflindholm, line 2: Could probably use a Qualcomm prefix on both this file and RamPartition.h.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+  Build HOB for the memory region
+
+  @param  MemRegion       pointer to the memory region
+
+  @retval EFI_SUCCESS     Successfully retrieves memory base and size
+  @retval EFI_INVALID_PARAMETER  The RAM partition table is invalid
+**/
+STATIC EFI_STATUS
+BuildMemoryHob (
+  IN MEM_REGION_INFO  *pMemRegion
```
leiflindholm: This file may be somewhat consistent with its Hungarian notation, but I still hate it.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+  return EFI_SUCCESS;
+}
+
+STATIC VOID
+AddMemRegionHobs (
+  VOID
+  )
+{
+  UINTN            Index;
+  MEM_REGION_INFO  *MemRegions   = NULL;
```
leiflindholm, line 138: Separate definition and initialization.
More instances further down in file.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+  UINTN            MemRegionsCnt = 0;
+
+  GetMemRegionCfgInfo (&MemRegions, &MemRegionsCnt);
+  if ((MemRegions == NULL) || (MemRegionsCnt == 0)) {
+    DEBUG ((DEBUG_ERROR, "UEFI Memory Map configuration not found\r\n"));
+    ASSERT (MemRegions != NULL);
+    ASSERT (MemRegionsCnt != 0);
+    CpuDeadLoop ();
+    return; /* For KW */
```
leiflindholm: What's a KW?

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+      .Attributes   = ARM_MEMORY_REGION_ATTRIBUTE_WRITE_BACK
+    },
+    {
+      .PhysicalBase = FixedPcdGet64 (PcdSmemBaseAddress),
+      .VirtualBase  = FixedPcdGet64 (PcdSmemBaseAddress),
+      .Length       = FixedPcdGet64 (PcdSmemSize),
+      .Attributes   = ARM_MEMORY_REGION_ATTRIBUTE_UNCACHED_UNBUFFERED
+    },
+    {
+      .PhysicalBase = FixedPcdGet64 (PcdIMemCookiesBase),
```
leiflindholm, line 76: This reminds me this file is missing a glossary.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+      case AddPeripheral:
+        BuildMemoryHob (&MemRegions[Index]);
+        break;
+
+      case HobOnlyNoCacheSetting:
+        BuildMemoryHob (&MemRegions[Index]);
+        break;
+
+      case NoBuildHob:
+        /* Don't do anything, only cache is initialized */
```
leiflindholm: Initializing a cache sounds like something.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+        break;
+
+      case NoMap:
+        /* Don't do anything */
```
leiflindholm: Yes, doing a 'break;' immediately means that. Does not need a comment. Same below.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+        break;
+
+      case ErrorBuildHob:
+      default:
+        DEBUG ((DEBUG_ERROR, "Invalid BuildHob Option\n"));
+        ASSERT (FALSE);
+        CpuDeadLoop ();
+        break;
+    } /* end of switch */
```
leiflindholm: Yes, that's how C works. Drop these two comments please.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+}
+
+#define MAX_MEMORY_ENTRIES  (128)
+
+EFI_STATUS
+EFIAPI
+GeneratePageTableRegionMap (
+  IN MEM_REGION_INFO               *pMemRegions,
+  IN UINTN                         RegionsCnt,
+  IN ARM_MEMORY_REGION_DESCRIPTOR  **VirtualMemoryMap
```
leiflindholm, line 203: Well, not that consistent, I guess...

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+  )
+{
+  STATIC ARM_MEMORY_REGION_DESCRIPTOR  MemoryTable[MAX_MEMORY_ENTRIES];
+  ARM_MEMORY_REGION_ATTRIBUTES         CacheAttributes;
+  UINTN                                MemRgnCnt;
+  UINTN                                MemoryTableIndex = 0;
```
leiflindholm: There's only one index in the function.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+EFIAPI
+GeneratePageTableRegionMap (
+  IN MEM_REGION_INFO               *pMemRegions,
+  IN UINTN                         RegionsCnt,
+  IN ARM_MEMORY_REGION_DESCRIPTOR  **VirtualMemoryMap
+  )
+{
+  STATIC ARM_MEMORY_REGION_DESCRIPTOR  MemoryTable[MAX_MEMORY_ENTRIES];
+  ARM_MEMORY_REGION_ATTRIBUTES         CacheAttributes;
+  UINTN                                MemRgnCnt;
```
leiflindholm: MemoryRegionCount. We can affort the storage.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+
+    if (pMemRegions[MemRgnCnt].CacheAttributes == ARM_MEMORY_REGION_ATTRIBUTE_WRITE_BACK) {
+      MemoryTable[MemoryTableIndex].Attributes = CacheAttributes;
+    } else {
+      MemoryTable[MemoryTableIndex].Attributes = pMemRegions[MemRgnCnt].CacheAttributes;
+    }
+
+    MemoryTableIndex++;
+
+    if (MemoryTableIndex >= MAX_MEMORY_ENTRIES) {
```
leiflindholm: We've already checked that RegionsCnt is less than MAX_MEMORY_ENTRIES, so presumably this is never true?

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+{
+  EFI_STATUS       Status         = EFI_UNSUPPORTED;
+  MEM_REGION_INFO  *mMemRegions   = NULL;
+  UINTN            mNumMemRegions = 0;
+  BOOLEAN          UsedStaticMap  = FALSE;
+
+  DEBUG ((DEBUG_INFO, "ArmPlatformGetVirtualMemoryMap\n"));
+
+  /* Reserved DDR Region for T32 CMM Script */
+  ArmPLatformSetupDebugBuffer ();
```
leiflindholm: PL -> Pl

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/PlatformConfiguration.c
```diff
@@ -0,0 +1,750 @@
...
+#include <PlatformConfiguration.h>
+
+#define MAX_MEMORY_REGIONS  125
+
+STATIC MEM_REGION_INFO  *mMemRegions        = NULL;
+STATIC UINTN            mNumMemRegions      = 0;
+STATIC UINTN            mNumMemoryMapRegion = 0;
+STATIC UINT64           mMemMapLow          = 0xFFFFFFFFFFFFFFFFULL;
+STATIC UINT64           mMemMapHigh         = 0;
+STATIC BOOLEAN          mIsMemMapHighNoMap  = FALSE;
```
leiflindholm, line 27: There's got to be something less confusing to name that.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/PlatformConfiguration.c
```diff
@@ -0,0 +1,750 @@
...
+  Memory = AllocatePages (EFI_SIZE_TO_PAGES (AllocationSize));
+  if (Memory != NULL) {
+    ZeroMem (Memory, AllocationSize);
+  }
+
+  return Memory;
+}
+
+/**
+  Allocate memory without freeing.
```
leiflindholm: Yes, memory is rarely freed by the act of allocating it.
Could we have a description of what the purpose of the function is instead?

Also, we seem to be mixing names of Alloc and Allocate in this file.
Please clean up and make consistent.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/PlatformConfiguration.c
```diff
@@ -0,0 +1,750 @@
...
+
+/**
+  Allocates and zeros a buffer.
+
+  @param  AllocationSize        The number of bytes to allocate and zero.
+
+  @return A pointer to the allocated buffer or NULL if allocation fails.
+
+**/
+VOID *EFIAPI
```
leiflindholm: Uncrustify please.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/PlatformConfiguration.c
```diff
@@ -0,0 +1,750 @@
...
+  )
+{
+  EFI_STATUS       Status;
+  MEM_REGION_INFO  RamPartTableEntry[RAM_NUM_PART_ENTRIES];
+  UINTN            RamPartTableEntryCount;
+  UINT8            NoOverlap;
+  UINT32           Index;
+  UINT32           RamPartitionIndex;
+
+  NoOverlap         = 0;
```
leiflindholm: Is this Number of Overlap or No Overlap?
If the latter, please make it BOOL and also initialize it to TRUE, since it cannot reasonably be indicating overlapping before we've checked anything.

But I suspect we could probably improve things by renaming it.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/PlatformConfiguration.c
```diff
@@ -0,0 +1,750 @@
...
+  Index             = 0;
```
leiflindholm: More redundant Index zeroing.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/PlatformConfiguration.c
```diff
@@ -0,0 +1,750 @@
...
+
+  RamPartTableEntryCount = RAM_NUM_PART_ENTRIES;
+  Status = GetRamPartitions (&RamPartTableEntryCount, &RamPartTableEntry[0]);
+  if (Status != EFI_SUCCESS) {
+    ASSERT (Status == EFI_SUCCESS);
+    return Status;
+  }
+
+  for (Index = 0; Index < mNumMemoryMapRegion; Index++) {
+    NoOverlap = 0;
```
leiflindholm: If we're initializing it on every pass of the loop, and only accessing it inside the loop, then:
1) Why is it defined at function scope?
2) Why is it initialized before the loop?

Platform/Qualcomm/QualcommPlatformPkg/Library/ReportCpuHobLib/ReportCpuHobLib.c
```diff
@@ -0,0 +1,29 @@
+/** @file
+  Source code file for Report CPU HOB library.
```
leiflindholm, line 2: Open question:
Is is worth pointing out when something is implemented specifically for MinPlatform support?

I think it probably should be in the commit message, but wouldn't hurt here either.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/AArch64/ArmPlatformHelper.S
```diff
@@ -0,0 +1,12 @@
+//
```
leiflindholm, line 1: Typo in commit message: ArmPltformLib

Platform/Qualcomm/QualcommPlatformPkg/Library/PeiReportFvLib/PeiReportFvLib.c
```diff
@@ -0,0 +1,298 @@
...
+#include <Library/BaseMemoryLib.h>
+#include <Library/DebugLib.h>
+#include <Library/HobLib.h>
+#include <Library/PeiServicesLib.h>
+#include <Library/ReportFvLib.h>
+#include <Guid/FirmwareFileSystem2.h>
+#include <Ppi/FirmwareVolumeInfo2.h>
+
+// Use a FV pointer PCD to get a pointer to the FileSystemGuid in the FV header
+#define PCD_TO_FV_HEADER_FILE_SYSTEM_GUID(Pcd)  (&((EFI_FIRMWARE_VOLUME_HEADER *)(UINTN) PcdGet32 (Pcd))->FileSystemGuid)
```
leiflindholm: Is a 32-bit pointer safe here?

Platform/Qualcomm/QualcommPlatformPkg/Library/PeiReportFvLib/PeiReportFvLib.c
```diff
@@ -0,0 +1,298 @@
...
+  Status     = EFI_SUCCESS;
+  FvHeader   = NULL;
+  BootMode   = BOOT_WITH_FULL_CONFIGURATION;
+
+  Status = PeiServicesGetBootMode (&BootMode);
+  ASSERT_EFI_ERROR (Status);
+
+  DEBUG_CODE (
+    for (Index = 0; Status == EFI_SUCCESS; Index++) {
+    Status = PeiServicesLocatePpi (&gEfiPeiFirmwareVolumeInfo2PpiGuid, Index, &Descriptor, (VOID **)&Ppi);
```
leiflindholm, line 42: indentation

Platform/Qualcomm/QualcommPlatformPkg/Library/PeiReportFvLib/PeiReportFvLib.c
```diff
@@ -0,0 +1,298 @@
...
+  )
+{
+  UINTN                              Index;
+  EFI_PEI_PPI_DESCRIPTOR             *Descriptor;
+  EFI_PEI_FIRMWARE_VOLUME_INFO2_PPI  *Ppi;
+  EFI_STATUS                         Status;
+  EFI_FIRMWARE_VOLUME_HEADER         *FvHeader;
+  EFI_BOOT_MODE                      BootMode;
+
+  Index      = 0;
```
leiflindholm: No

Platform/Qualcomm/QualcommPlatformPkg/Library/PeiReportFvLib/PeiReportFvLib.c
```diff
@@ -0,0 +1,298 @@
...
+  Descriptor = NULL;
+  Ppi        = NULL;
+  Status     = EFI_SUCCESS;
```
leiflindholm: No.

Platform/Qualcomm/QualcommPlatformPkg/Library/PeiReportFvLib/PeiReportFvLib.c
```diff
@@ -0,0 +1,298 @@
...
+  VOID
+  )
+{
+  UINTN                              Index;
+  EFI_PEI_PPI_DESCRIPTOR             *Descriptor;
+  EFI_PEI_FIRMWARE_VOLUME_INFO2_PPI  *Ppi;
+  EFI_STATUS                         Status;
+  EFI_FIRMWARE_VOLUME_HEADER         *FvHeader;
+
+  Index      = 0;
```
leiflindholm: I spy another antipattern.
Please go through and nuke all redundant initializations.

Platform/Qualcomm/QualcommPlatformPkg/Library/PeiReportFvLib/PeiReportFvLib.inf
```diff
@@ -0,0 +1,63 @@
...
+#
+# Copyright (c) 2018 - 2021, Intel Corporation. All rights reserved.<BR>
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.<BR>
+#
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+##
+
+[Defines]
+  INF_VERSION                    = 0x00010017
```
leiflindholm: *cough*

Platform/Qualcomm/QualcommPlatformPkg/QualcommPlatformPkg.dsc.inc
```diff
@@ -0,0 +1,69 @@
...
+  MdeModulePkg/Bus/Pci/SataControllerDxe/SataControllerDxe.inf
+  MdeModulePkg/Bus/Ata/AtaAtapiPassThru/AtaAtapiPassThru.inf
+  MdeModulePkg/Bus/Ata/AtaBusDxe/AtaBusDxe.inf
+  MdeModulePkg/Bus/Pci/NvmExpressDxe/NvmExpressDxe.inf
+  MdeModulePkg/Universal/DevicePathDxe/DevicePathDxe.inf
+  MdeModulePkg/Universal/Disk/DiskIoDxe/DiskIoDxe.inf
+  MdeModulePkg/Universal/Disk/PartitionDxe/PartitionDxe.inf
+
+[BuildOptions]
+  *_*_AARCH64_CC_FLAGS = -fno-omit-frame-pointer -fno-optimize-sibling-calls -fno-inline
```
leiflindholm: Err, why? The latter two of those look like hiding bugs, and they're presumably going to have performance impact in RELEASE builds.

leiflindholm (dismissed):

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,373 @@
...
+  }
+
+  Status = InitRamPartitionTableLib ();
+  if (Status != EFI_SUCCESS) {
+    DEBUG ((DEBUG_ERROR, "InitRamPartitionTableLib failed\n"));
+    goto ExitError;
+  }
+
+  /* Try DT-based platform configuration first */
+  Status = LoadPlatformConfigFromDeviceTree ();
```
b49020, line 514: What kind of Devicetree support is this? Is it the boot DT or the kernel DT?

b49020:
@vishalo this PR is adding a lot of Qcom specific ABI interfaces among edk2 and the earlier stage boot-loaders like XBL. The commit messages are lacking any description of what those ABI interfaces are. Can you rather add some documentation alongside for these ABI interfaces? As they have to be maintained across multiple targets.

Also, please try to keep in mind that we are already supporting open source early stage boot-loaders like TF-A BL2 chainloaded by XBL. So better we limit any ABI overhead or make these ABIs optional like for QHEE or QSEE.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
vishalo: boot DT.

vishalo:
> @vishalo this PR is adding a lot of Qcom specific ABI interfaces among edk2 and the earlier stage boot-loaders like XBL. The commit messages are lacking any description of what those ABI interfaces are. Can you rather add some documentation alongside for these ABI interfaces? As they have to be maintained across multiple targets.

Agree on importance of documentation - what is the right place for this documentation, and what is the datail? Give this is produced by earlier stage, shouldn't it be in the producer side?
> 
> Also, please try to keep in mind that we are already supporting open source early stage boot-loaders like TF-A BL2 chainloaded by XBL. So better we limit any ABI overhead or make these ABIs optional like for QHEE or QSEE.

How is TFA-BL2 handling this? Is this documented in TFA-BL2?

> So better we limit any ABI overhead or make these ABIs optional like for QHEE or QSEE.
What do you mean by make these ABIs optional ? You mean have UEFI not depend on it ?

b49020:
> Agree on importance of documentation - what is the right place for this documentation, and what is the datail? Give this is produced by earlier stage, shouldn't it be in the producer side?

The producer side is all downstream here being the XBL, QHEE or QSEE. So on consumer side at-least we can document the ABI. Feel free to add documentation under `Platform/Qualcomm` to begin with as otherwise the custom ABIs can become hard to maintain upstream.

> How is TFA-BL2 handling this? Is this documented in TFA-BL2?

Right now for Qcom platforms, we don't have any ABI with TF-A BL2. Going forwards we should just adopt the standard Firmware Handoff spec [1] as all other silicon vendors do.

[1] https://firmwarehandoff.github.io/firmware_handoff/main/index.html

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
b49020: AFAIK, upstream edk2 has avoided supporting other DTs apart from kernel DTs. Can we not avoid special boot DTs here? I see you already are adding support for SMEM and IMEM based ABI mechanisms.

leiflindholm: I can't parse this comment.
If it's not for kernel consumption, why should anyone care?

b49020: Yeah having a custom DT just for Qcom boot stack to pass information around will be complicated to maintain the ABI unless you make a standard DT bindings for yourselves. As you know how forwards and backwards compatibility with DT works. People often tend to break ABIs.

leiflindholm: There is no ABI other than what Qualcomm defines it to be. And the only consumer is other Qualcomm code.
DT is a storage format. It is only more complicated than other storage formats because people get confused and think it's something that the kernel owns. It's not.

b49020: > There is no ABI other than what Qualcomm defines it to be. And the only consumer is other Qualcomm code.

Don't we require a DT binding there that such an ABI doesn't break since we have to be forward and backwards compatible with whatever runs as early stage boot-loader?

> DT is a storage format. It is only more complicated than other storage formats because people get confused and think it's something that the kernel owns. It's not.

I am not concerned about ownership here but rather how a common understanding is there for DT to describe hardware rather than being used as a software construct. Although every other project tends to start using it as a software construct. Not sure if that is true with ACPI or not.

leiflindholm: What project? This is a qualcomm-internal interface. This could be XML or static C structs for what anyone should care, but for some reason, as soon as some poor sod makes the decision to reuse libfdt because it's already in the project, people pop up thinking because the same storage format was used for something specific somewhere else, the same rules should apply anywhere. This never happens with any other format.

b49020: > What project?

TF-A does already use DT as software construct too but it's mostly for internal consumption, I am sure there will be many other examples. However, passing custom DT as a custom software construct across boot stages is something I was concerned.

Anyhow fair enough, I can live with another usage of DT as a software construct if the ABI is documented properly. Surely I hope with the open source boot stages like TF-A BL2 we are just going to limit ABI to whatever other silicon vendors have adopted as Firmware Handoff spec.

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
```diff
@@ -0,0 +1,145 @@
...
+    that a delayed ramdump backup is needed.
+    Uses 0x44554d50 instead of legacy 0xC1F8DB42 to enable dumping
+    multiple TZ segments rather than a legacy TZ binary blob. */
+#define QUALCOMM_QHEE_DLOAD_DUMP_SHARED_IMEM_MAGIC_NUM  0x44554d50
+#define QUALCOMM_QHEE_DLOAD_DUMP_SHARED_IMEM_OFFSET     0xb20
+
+#define QUALCOMM_BOOT_DUMP_NUM_ENTRIES    6
+#define QUALCOMM_DLOAD_DUMP_INFO_VERSION  1
+
+#ifdef FEATURE_TPM_HASH_POPULATE
```
b49020: Is this `ifdef` really needed? I don't see it being enabled anywhere.

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
```diff
@@ -0,0 +1,145 @@
...
+    that a delayed ramdump backup is needed. */
+#define QUALCOMM_QSEE_DLOAD_DUMP_SHARED_IMEM_OFFSET     0x748
+#define QUALCOMM_QSEE_DLOAD_DUMP_SHARED_IMEM_MAGIC_NUM  0x44554d50
+
+/** Shared IMEM offset and magic number written by QHEE to signal XBL
+    that a delayed ramdump backup is needed.
+    Uses 0x44554d50 instead of legacy 0xC1F8DB42 to enable dumping
+    multiple TZ segments rather than a legacy TZ binary blob. */
+#define QUALCOMM_QHEE_DLOAD_DUMP_SHARED_IMEM_MAGIC_NUM  0x44554d50
+#define QUALCOMM_QHEE_DLOAD_DUMP_SHARED_IMEM_OFFSET     0xb20
```
b49020: Are these QSEE and QHEE macros of any use to edk2? As per comments these are used by XBL to consume data?

Platform/Qualcomm/QualcommPlatformPkg/Include/Library/RamPartitionTableLib.h
b49020, line 1: AFAICS, the only APIs which are consumed by ArmPlatformLib from this library are:

```
RamPartitionInitRamPartitionTableLib()
RamPartitionGetRamPartitions()
```

I don't see rest of the APIs used externally. If they are used internally then make them static or just drop any APIs which don't have a consumer.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,372 @@
...
+    {
+      .PhysicalBase = FixedPcdGet64 (PcdIMemCookiesBase),
+      .VirtualBase  = FixedPcdGet64 (PcdIMemCookiesBase),
+      .Length       = FixedPcdGet64 (PcdIMemCookiesSize),
+      .Attributes   = ARM_MEMORY_REGION_ATTRIBUTE_DEVICE
+    },
+    { 0 } // End of table
+  };
+
+  return ArmConfigureMmu (EarlyInitMemoryTable, NULL, NULL);
```
b49020, line 84: How much gain does this early MMU enable gives us? Is it majorly all the DT parsing benefits from this?

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -0,0 +1,372 @@
...
+  VOID
+  )
+{
+  ARM_MEMORY_REGION_DESCRIPTOR  EarlyInitMemoryTable[] = {
+    {
+      .PhysicalBase = FixedPcdGet64 (PcdSerialRegisterBase),
+      .VirtualBase  = FixedPcdGet64 (PcdSerialRegisterBase),
+      .Length       = EFI_PAGE_SIZE,
+      .Attributes   = ARM_MEMORY_REGION_ATTRIBUTE_DEVICE
+    },
```
b49020, line 56: Why don't we map the entire device region in one go? This peripheral specific MMU mapping is expensive requiring all the page level mappings.

b49020 (changes requested):

Platform/Qualcomm/QualcommPlatformPkg/Include/Library/RamPartitionTableLib.h
vishalo: They will be used when we enable more features.

b49020: Yeah so please add them when there is a consumer.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
vishalo: Then we are not using the capabilities of MMU. Writes to invalid addresses in UEFI will end up as NOC errors ending up in secure side, and harder to debug. This helps catch the problem where/when it occurred.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
vishalo: Yes, it makes it much faster. I did not profile it, but there is notable delay.

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
vishalo: You're right, will remove.

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
vishalo: We can remove the ifdef ...

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
b49020: Okay I see your point. I suppose that's why you use DT for MMU setup for all the periperals.

Platform/Qualcomm/QualcommPlatformPkg/Include/Library/RamPartitionTableLib.h
vishalo: Added logs that print this information, since it is useful for triage.

b49020 (approved):
Looks good to me

Platform/Qualcomm/QualcommPlatformPkg/Include/BootSharedImemCookie.h
vishalo: For now I would like to keep this. If this doesn't make sense, I can look at getting it update.

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.c
vishalo: I would like to do this in another clean up PR.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
vishalo: KlockWork, , removed.

Platform/Qualcomm/QualcommPlatformPkg/Library/ReportCpuHobLib/ReportCpuHobLib.c
vishalo: Can be used by non minplatform as well.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLib.inf
```diff
@@ -0,0 +1,72 @@
...
+
+[Packages]
+  ArmPkg/ArmPkg.dec
+  ArmPlatformPkg/ArmPlatformPkg.dec
+  EmbeddedPkg/EmbeddedPkg.dec
+  MdeModulePkg/MdeModulePkg.dec
+  MdePkg/MdePkg.dec
+  QcomPlatformPkg/QcomPlatformPkg.dec
+  QualcommPlatformPkg/QualcommPlatformPkg.dec
+  QualcommSiliconPkg/QualcommSiliconPkg.dec
```
b49020, line 31: The build fails for me here, I need following changes to proceed:

```
-  QcomPlatformPkg/QcomPlatformPkg.dec
-  QualcommPlatformPkg/QualcommPlatformPkg.dec
-  QualcommSiliconPkg/QualcommSiliconPkg.dec
+  Platform/Qualcomm/QualcommPlatformPkg/QualcommPlatformPkg.dec
+  Silicon/Qualcomm/QualcommSiliconPkg/QualcommSiliconPkg.dec
```

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.inf
```diff
@@ -0,0 +1,41 @@
...
+#  VALID_ARCHITECTURES           = AARCH64
+#
+
+[Sources]
+  RamPartitionTableLib.c
+
+[Packages]
+  ArmPkg/ArmPkg.dec
+  QualcommPlatformPkg/QualcommPlatformPkg.dec
+  QualcommSiliconPkg/QualcommSiliconPkg.dec
```
b49020, line 31: Ditto here:

```
-  QualcommPlatformPkg/QualcommPlatformPkg.dec
-  QualcommSiliconPkg/QualcommSiliconPkg.dec
+  Platform/Qualcomm/QualcommPlatformPkg/QualcommPlatformPkg.dec
+  Silicon/Qualcomm/QualcommSiliconPkg/QualcommSiliconPkg.dec
```

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLib.inf
vishalo: Are these for building current patch set or downstream changes ? Are you using the  scripts to build or invoking build directly passing dsc ? Likely PACKAGES_PATH is not set correctly. Adding "Platform/Qualcomm" is not the convention ..

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.inf
vishalo: Are these for building current patch set or downstream changes ? Looks like PACKAGES_PATH is not set when you invoke build.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLib.inf
b49020: > Adding "Platform/Qualcomm" is not the convention ..

That's right, I am just using edk2 and edk2-platforms in the `PACKAGES_PATH`. This is the same practice that build systems like Yocto recipes uses. Let's try to avoid this extra `PACKAGES_PATH` requirement.

Platform/Qualcomm/QualcommPlatformPkg/Library/RamPartitionTableLib/RamPartitionTableLib.inf
b49020: I am trying to come up with a MinPlatform upstream port for Nord IQ10 EVK reusing the infrastructure you are adding here.

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLib.inf
vishalo: There will be more paths we need to add as we get more functionality enabled.

b49020: Okay, let me merge this PR as it is. Surely we need to limit nested directories in the `PACKAGES_PATH`. Let's take that as a follow-up work item.

PR #1006, opened 2026-07-02
MinPlatformPkg: Update FADT/FACS to indicate 6.6 support

lakhuanx:
Update FADT/FACS to indicate 6.6 support


SaiChaganty (approved):

SaiChaganty (commented):
on a second thought, do we really need to update the version, in the .dec file? 
Can we handle the version update via an override, in a platform .dsc file?

abdattar (approved):

PR #1013, opened 2026-07-14
IntelSiliconPkg: Add CONFIG_BLOCK_HEADER2 with multi-instance support

nate-desimone:
## Description

Introduce `CONFIG_BLOCK_HEADER2`, a layout-compatible superset of `CONFIG_BLOCK_HEADER` that appends a `Namespace` GUID at offset 32. The `Namespace` disambiguates multiple config blocks that share the same `GuidHob.Name`, enabling platforms with more than one instance of a given config block schema to coexist in the same config block table.

`CONFIG_BLOCK_HEADER2_ATTRIBUTE` (`BIT0` of the `Attributes` field) is the explicit format discriminator.  `AddConfigBlock()` is updated to copy the full extended header when this bit is set, existing users of
`CONFIG_BLOCK_HEADER` are unaffected.

Two new library APIs are added to `ConfigBlockLib`:
  - `GetConfigBlockByInstance()`: look up a block by type GUID and `Namespace` GUID.
  - `GetNextConfigBlock()`: iterate blocks in the table, with an optional filter by type GUID.

### DIFF
--- a/Silicon/Intel/IntelSiliconPkg/Library/BaseConfigBlockLib/BaseConfigBlockLib.c
+++ b/Silicon/Intel/IntelSiliconPkg/Library/BaseConfigBlockLib/BaseConfigBlockLib.c
@@ -1,7 +1,7 @@
 /** @file
   Library functions for Config Block management.
 
-Copyright (c) 2017 - 2019, Intel Corporation. All rights reserved.<BR>
+Copyright (c) 2017 - 2026, Intel Corporation. All rights reserved.<BR>
 SPDX-License-Identifier: BSD-2-Clause-Patent
 
 **/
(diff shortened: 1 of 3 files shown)


Silicon/Intel/IntelSiliconPkg/Library/BaseConfigBlockLib/BaseConfigBlockLib.c
```diff
@@ -84,7 +84,20 @@ AddConfigBlock (
...
 
   TempConfigBlk = (CONFIG_BLOCK *)((UINTN)ConfigBlkTblAddrPtr + (UINTN)(ConfigBlkTblAddrPtr->Header.GuidHob.Header.HobLength - ConfigBlkTblAddrPtr->AvailableSize));
-  CopyMem (&TempConfigBlk->Header, &ConfigBlkAddrPtr->Header, sizeof(CONFIG_BLOCK_HEADER));
+
+  //
+  // If the template carries CONFIG_BLOCK_HEADER2 format, copy the full version 2
+  // header (including the Namespace GUID at offset 32); otherwise copy only the
+  // base CONFIG_BLOCK_HEADER.
+  //
+  if ((ConfigBlkAddrPtr->Header.Attributes & CONFIG_BLOCK_HEADER2_ATTRIBUTE) != 0) {
```
SaiChaganty, line 93: Perhaps it would be safe to check for BIT0 instead of !=0 ?

PR #1029, opened 2026-07-28
Qualcomm/Common: Add ChipInfoLib and PlatformInfoLib support

vishalo:
Add ChipInfoLib and PlatformInfoLib support for querying chip and platform information.

Add support to log this during platform initialization.

Common/* contains code that can be shared between MinPlatform and non-MinPlatform implemtentation.


Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
+/** @file
+  This is the interface specification for the chip driver/service.
+
+  @par Glossary:
```
leiflindholm: I think the glossary should be below Copyright/license. (Throughout.)

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/QualcommCommonSiliconPkg.dec
```diff
@@ -0,0 +1,26 @@
...
+# Include Section - list of Include Paths that are provided by this package.
+#                   Comments are used for Keywords and Module Types.
+#
+################################################################################
+
+[Includes.common]
+  Include
+
+[Guids]
+  gQualcommCommonSiliconPkgTokenSpaceGuid     = {0xDCC48FA6, 0X5910, 0X4A76, {0X90, 0X19, 0X7D, 0X67, 0X67, 0X21, 0XF3, 0XC2}}
```
leiflindholm: `0x` please, not `0X`.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+
+/**
+  Returns the version of the chip as read from fuses.
+
+  This raw version is a strictly increasing integer counter incremented for
+  each new HW version (e.g. v1.0 = 0, v1.1 = 1, v2.0 = 3, ...).
+
+  The only use case for this API is for patch revisions, i.e., the 'z' in
+  vX.Y.Z. Patch revisions are not reflected in the value returned from
+  ChipInfoGetChipVersion, but are included in the HW revision counter read
```
leiflindholm: This smells of LLM failing to ignore the fact that it always has the full context of the program, making it a useless description and bound to be incorrect in the future.

Document what the function logically _does_, not how you expect it to be implemented.

The description is also really confusing wrt difference between "raw version" and "version". Is this minor/major version?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/QualcommCommonSiliconPkg.dec
b49020, line 1: Why do we need a new package here? Why don't you reuse https://github.com/tianocore/edk2-platforms/tree/master/Silicon/Qualcomm/QualcommSiliconPkg?

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/QualcommCommonPlatformPkg.dec
b49020, line 1: Ditto, what's wrong with reuse of https://github.com/tianocore/edk2-platforms/tree/master/Platform/Qualcomm/QualcommPlatformPkg?

b49020 (dismissed):

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+EFIAPI
+ChipInfoGetChipIdString (
+  OUT CHAR8   *ChipIdStr,
+  IN  UINT32  MaxLength
+  );
+
+/**
+  Returns the chip's family as a ChipInfoFamilyType.
+
+  This has no connection to the architectural family - it's used to group
```
leiflindholm: This is a very confusing statement.
If we've managed to confuse our internal terminology, we can still give the external API a less confusing name.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+           or if called before ChipInfo is initialized.
+**/
+ChipInfoFamilyType
+EFIAPI
+ChipInfoGetChipFamily (
+  VOID
+  );
+
+/**
+  Returns the marketing name of the processor.
```
leiflindholm: "product"
That it happens to come from the marketing department isn't something we ought to leak into the API.
(Also affects function/argument names and some more of the comment block.)

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+ChipInfoGetRawChipId (
+  VOID
+  );
+
+/**
+  Returns the name of the chip (e.g. MSM8998, MDM9665).
+
+  @param[out]  ChipIdStr  Pointer to a buffer to hold the chip's name. If
+                           called before ChipInfo is initialized, or if the
+                           chip ID string is unknown, "UNKNOWN" will be
```
leiflindholm: "will be copied" is passive voice.
'the function sets the buffer to "UNKNOWN"'
(More instances in this file.)

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+EFIAPI
+ChipInfoGetMarketingNameString (
+  OUT CHAR8   *MarketingNameStr,
+  IN  UINT32  MaxLength
+  );
+
+/**
+  Returns whether or not the chip supports a modem.
+
+  The value returned should be treated as indicating the presence of a modem:
```
leiflindholm: LLM smell.
Redundant (and passive), the return description is already complete.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+EFIAPI
+ChipInfoGetSerialNumber (
+  VOID
+  );
+
+/**
+  Returns the foundry ID for the chip as a ChipInfoFoundryIdType.
+
+  @return  Foundry ID if successful, or
+           CHIPINFO_FOUNDRYID_UNKNOWN if called before ChipInfo is initialized.
```
leiflindholm: Should be FOUNDRY_ID (both logically and to match function name).

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+**/
+ChipInfoFoundryIdType
+EFIAPI
+ChipInfoGetFoundryId (
+  VOID
+  );
+
+/**
+  Returns the chip's architecture family number as read from HW.
```
leiflindholm, line 167: What is an architecture family? We already have confusing statements in this file about what family means.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+  VOID
+  );
+
+/**
+  Returns the chip's QFPROM chip ID.
+
+  Can be used along with the serial number to uniquely identify the chip.
+
+  @return  Chip ID as read from QFPROM fuses if successful, or
+           CHIPINFO_QFPROM_CHIPID_UNKNOWN if called before ChipInfo is
```
leiflindholm: CHIP_ID

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+**/
+ChipInfoQfpromChipIdType
+EFIAPI
+ChipInfoGetQfpromChipId (
+  VOID
+  );
+
+/**
+  Retrieves a mask of CPUs in the specified cluster marked as disabled in
+  PTE fuses. If the cluster has fewer than 32 CPUs, the additional bits are
```
leiflindholm: "are treated": passive, and in this situation actually confusing. Please rewrite.
e.g.
"Bits corresponding to CPUs not implemented indicate 'disabled'."?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+EFIAPI
+ChipInfoGetDisabledCpus (
+  IN  UINT32  CpuCluster,
+  OUT UINT32  *Mask
+  );
+
+/**
+  Initialize the ChipInfo driver.
+
+  Any missing or unknown information will be indicated as specified
```
leiflindholm: What does this sentence mean?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+  @return  CHIPINFO_SUCCESS always.
+**/
+ChipInfoResult
+EFIAPI
+ChipInfoInit (
+  VOID
+  );
+
+/**
+  Gets SKU and product code information for the current device.
```
leiflindholm: Previous descriptions in this file uses "returns", which is better. Next few comment blocks also use "gets". Please use "returns" throughout.

Unrelated but same line:
What device? SoC?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+
+  This information can be used to make functional decisions, e.g.,
```
leiflindholm: Not relevant.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+  @param[out]  ChipIdStr  Pointer to a buffer to hold the chip's name. If
+                           called before ChipInfo is initialized, or if the
+                           chip ID string is unknown, "UNKNOWN" will be
+                           copied into this buffer.
+  @param[in]   MaxLength  Length of the string to copy. Max supported size
+                           is CHIPINFO_MAX_ID_LENGTH.
+
+  @return  CHIPINFO_SUCCESS always.
+**/
+ChipInfoResult
```
leiflindholm: Hmm...

Why are we using custom return types?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+                      information will be stored.
+
+  @return  CHIPINFO_SUCCESS if successful.
+  @return  CHIPINFO_ERROR_INVALID_PARAMETER if Info is NULL.
+  @return  CHIPINFO_ERROR_NOT_SUPPORTED if this target does not support
+           SKU information.
+**/
+ChipInfoResult
+EFIAPI
+ChipInfoGetSKU (
```
leiflindholm: In CamelCase, SKU should be written as Sku (throughout).

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+ChipInfoResult
+EFIAPI
+ChipInfoGetNumFunctionalClusters (
+  OUT UINT32  *NumClusters
+  );
+
+/**
+  Gets the boot cluster and core indices.
+
+  This function returns an error type instead of returning the queried
```
leiflindholm: Not really relevant, and if it was, this could be described a whole lot less verbosely.

Ultimately, the use of "core" and "cluster" ties this to the Arm architecture, so could possibly reference format being identical to corresponding MPIDR_EL1 AFFx field, and drop the zero-indexed note below.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+EFIAPI
+ChipInfoGetBootClusterAndCore (
+  OUT UINT32  *Cluster,
+  OUT UINT32  *Core
+  );
+
+/**
+  Gets the HW part-level disabled feature details from fuses.
+
+  The bit-level interpretation of Mask is specific to the client.
```
leiflindholm: Wat?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+
+  @param[in]   Part  The ChipInfoPartType being queried.
+  @param[in]   Idx   Hardware instance of the selected part.
+  @param[out]  Mask  Used to store the part-specific details read from the
```
leiflindholm: "read from the QTV/PTE region" belongs in the function description, not the parameter description.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+                      QTV/PTE region.
+
+  @return  CHIPINFO_SUCCESS if successful.
+  @return  CHIPINFO_ERROR_INVALID_PARAMETER if Mask is NULL or Part is unknown.
+  @return  CHIPINFO_ERROR_OUT_OF_RANGE_PARAMETER if Part is out of range.
```
leiflindholm: s/out of range/invalid/?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,399 @@
...
+  TRUE indicates the part is completely disabled.
+  Use ChipInfoGetDisabledFeatures to obtain partial disable information.
+
+  @param[in]   Part  The ChipInfoPartType being queried.
+  @param[in]   Idx   Hardware instance of the selected part.
+  @param[out]  Mask  Pointer to a BOOLEAN to store the disabled status.
+
+  @return  CHIPINFO_SUCCESS if successful.
+  @return  CHIPINFO_ERROR_INVALID_PARAMETER if Mask is NULL or Part is unknown.
+  @return  CHIPINFO_ERROR_OUT_OF_RANGE_PARAMETER if Part is out of range.
```
leiflindholm: Wait? Why are we separating "part unknown" and "part out of range"? When is it useful to determine a part is unknown but "in range"?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/ChipInfoDefs.h
```diff
@@ -0,0 +1,1249 @@
...
+/*======================================================================
+ Includes
+ =====================================================================*/
+#include <Base.h>
+
+/*======================================================================
+ Macros
+ =====================================================================*/
+
+/**
```
leiflindholm: Doxygen tagging in this file is ... if not wrong, at least unusual.
Does this one have some meaning I'm not familiar with?

Below I also see blocks starting `/**`, but ending with `*/` instead of `**/`.
Is that valid?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/ChipInfoDefs.h
```diff
@@ -0,0 +1,1249 @@
...
+ @par Glossary:
+   - Qfprom - Qualcomm Fuse Programmable Read-Only Memory
+
+ Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.<BR>
+ SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#pragma once
+
+/*======================================================================
```
leiflindholm: Drop please, as well as the Includes one below, oh and further "section headers" further down in the file.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/ChipInfoDefs.h
```diff
@@ -0,0 +1,1249 @@
...
+ * Can be used by
+ * clients when declaring space for
+ * ChipInfoGetMarketingNameString.
+ */
+#define CHIPINFO_MAX_NAME_LENGTH  CHIPINFO_MAX_ID_LENGTH
+
+/**
+ * Maximum number of CPU clusters supported by the GetDefectiveParts API
+ */
+#define CHIPINFO_MAX_CPU_CLUSTERS  4
```
leiflindholm: The feels ... low?
Does the comment need expanding to clarify why?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/ChipInfoDefs.h
```diff
@@ -0,0 +1,1249 @@
...
+ *   ChipInfoGetChipVersion(&version);
+ *   if (version >= CHIPINFO_VERSION(1,2)) ...
+ */
+#define CHIPINFO_VERSION(major, minor)  (((major) << 16) | (minor))
+#define CHIPINFO_VERSION_UNKNOWN  0
+
+/**
+ * Maximum length of a chip id string.
+ *
+ * Can be used by clients when
```
leiflindholm: Getting that LLM smell again.
Who we expect to use a function, type, or macro is not useful documentation unless we're doing something ASTONISHINGLY esoteric.

Please do a global search for "client" and rewrite whatever comment blocks it's in that doesn't implement a client/server relationship. It's a nonsensical term outside of that context and a noticeable indicator of bad comments so far.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/ChipInfoDefs.h
```diff
@@ -0,0 +1,1249 @@
...
+#define CHIPINFO_MODEM_EVDO_REVB  0x0040
+#define CHIPINFO_MODEM_LTE        0x0080
+#define CHIPINFO_MODEM_GNSS       0x0100
+#define CHIPINFO_MODEM_EHRPD      0x0200
+#define CHIPINFO_MODEM_SUPPORTED  0xFFFFFFFF
+
+/*
+ * Define some modem combination shortcuts.
+ * These macros are deprecated, and haven't meant anything for several
+ * years. New development should only use CHIPINFO_MODEM_SUPPORTED or
```
leiflindholm: Then do we need to include them here?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/ChipInfoDefs.h
```diff
@@ -0,0 +1,1249 @@
...
+ */
+
+/**
+ * Chip version type with major number in the upper 16 bits and minor
+ * number in the lower 16 bits.
+ *
+ * For example:
+ *   1.0 -> 0x00010000
+ *   2.3 -> 0x00020003
+ * The CHIPINFO_VERSION macro should be used to generate the appropriate
```
leiflindholm: "Use the CHIPINFO_VERSION macro to..."

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/ChipInfoDefs.h
```diff
@@ -0,0 +1,1249 @@
...
+  CHIPINFO_FAMILY_MSM8630     = CHIPINFO_FAMILY_MSM8930,
+  CHIPINFO_FAMILY_MSM8230     = CHIPINFO_FAMILY_MSM8930,
+  CHIPINFO_FAMILY_APQ8030     = CHIPINFO_FAMILY_MSM8930,
+  CHIPINFO_FAMILY_MSM8627     = 30,
+  CHIPINFO_FAMILY_MSM8227     = CHIPINFO_FAMILY_MSM8627,
+  CHIPINFO_FAMILY_MSM8974     = 32,
+  CHIPINFO_FAMILY_MSM8625     = 33,
+  CHIPINFO_FAMILY_MSM8225     = CHIPINFO_FAMILY_MSM8625,
+  CHIPINFO_FAMILY_APQ8064     = 34,
+  CHIPINFO_FAMILY_MDM9x25     = 35,
```
leiflindholm: The opposite here, please use `X` instead of `x`. (As you can see it messes up the syntax highlighting, but it does also break the coding style.)

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/ChipInfoDefs.h
```diff
@@ -0,0 +1,1249 @@
...
+  CHIPINFO_P_CODE_32BITS = 0x7fffffff
+} ChipInfoPCodeType;
+
+/**
+ * FeatureCode and P-Code information for the current device.
+ *
+ * See ChipInfoGetSKU for more details.
+ */
+typedef struct {
+  ChipInfoFeatureCodeType    eFeatureCode;
```
leiflindholm: Are we back in Hungary?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/ChipInfoDefs.h
```diff
@@ -0,0 +1,1249 @@
...
+  ChipInfoPCodeType          ePCode;
+} ChipInfoSKUType;
+
+/**
+ @}
+ *//* end_group Data Structures */
```
leiflindholm: Drop end comment.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Library/ChipInfoLib/ChipInfo.c
```diff
@@ -0,0 +1,587 @@
...
+ Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.<BR>
+ SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+/*=========================================================================
+ Include Files
+ ==========================================================================*/
+
+#include <Base.h>
+#include "ChipInfoDefs.h"
```
leiflindholm: Umm, this ought to be `<Library/ChipInfoDefs.h>`?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Library/ChipInfoLib/ChipInfo.c
```diff
@@ -0,0 +1,587 @@
...
+ This file implements the Chip Info APIs.
+
+ @par Glossary:
+ - Qfprom - Qualcomm Fuse Programmable Read-Only Memory
+
+ Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.<BR>
+ SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+/*=========================================================================
```
leiflindholm: Right, global search and destroy on "and now we're in this part of the file" comment blocks please - this isn't pascal.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Library/ChipInfoLib/ChipInfoLib.c
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+  Internal structure to hold chip info.
+
+  During XBL Core, ChipInfoDxe init pulls data from the SMEM region
+  and stores it in here.
+
+  The query APIs read data from this structure.
+**/
+ChipInfoCtxtType  mChipInfoCtxt;
```
leiflindholm: `Ctx` is in the codebase as an abbreviated version of Context, `Ctxt` is not. Could you align please?

Both type and variable names.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Library/ChipInfoLib/ChipInfo.c
```diff
@@ -0,0 +1,587 @@
...
+ChipInfoResult
+EFIAPI
+ChipInfoGetChipIdString (
+  OUT CHAR8   *ChipIdStr,
+  IN  UINT32  MaxLength
+  )
+{
+  UINT32  Len;
+
+  Len = 0;
```
leiflindholm: There is no path on which Len is not initialized. Please drop.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Library/ChipInfoLib/ChipInfo.c
```diff
@@ -0,0 +1,587 @@
...
+ChipInfoGetDisabledCpus (
+  IN  UINT32  CpuCluster,
+  OUT UINT32  *Mask
+  )
+{
+  if (Mask == NULL) {
+    return CHIPINFO_ERROR;
+  }
+
+  if ((CpuCluster == 0) && !mChipInfoCtxt.NumClusters) {
```
leiflindholm: Isn't the `(CpuCluster == 0)` test redundant (and potentially incorrect)?
If NumClusters is 0 we always need to do this, right?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Library/ChipInfoLib/ChipInfo.c
```diff
@@ -0,0 +1,587 @@
...
+    //
+    // If this target doesn't have a binning plan, we don't know how many
+    // CPU clusters there are for this target, and ChipInfo.NumClusters
+    // will be left at its default value of 0. This is the only case where
+    // NumClusters will be 0.
+    //
+    // Some clients call this function even if a binning plan isn't defined,
```
leiflindholm: We don't care who calls us.
'When this happens, return success with Mask = "enabled/unknown" since there's always at least 1 core that's working (the one that's running this function).'

Can replace the rest of the comment block from here.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Library/ChipInfoLib/ChipInfoImage.c
```diff
@@ -0,0 +1,95 @@
...
+  Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.<BR>
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#include <Uefi.h>
+#include <Base.h>
+#include <Library/BaseLib.h>
+#include <Library/SmemLib.h>
+#include "ChipInfoImage.h"
+#include "PlatformInfoDefs.h"
```
leiflindholm: This is not a local include, use `<>`.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Library/ChipInfoLib/ChipInfoLib.c
```diff
@@ -0,0 +1,171 @@
...
+    // This will only happen in pre-sil on a standalone build.
+    // Return SUCCESS using the UNKNOWN values for each field,
+    // and let the information be set manually. However, mark the driver as
+    // initialized, to record that this has happened.
+    AsciiStrCpyS (mChipInfoCtxt.ChipIdString, CHIPINFO_MAX_ID_LENGTH, "UNKNOWN");
+    mChipInfoCtxt.InitComplete = TRUE;
+    return CHIPINFO_SUCCESS;
+  }
+
+  // Read the chip information from Smem
```
leiflindholm: Can you break into a helper function from here...

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Library/ChipInfoLib/ChipInfoLib.c
```diff
@@ -0,0 +1,171 @@
...
+  if (Smem->Format >= PLATFORMINFO_VERSION (0, 23)) {
+    Size                    = Smem->NumPartInstances * sizeof (PlatformInfoPartInfoType);
+    mChipInfoCtxt.PartInfo = AllocatePages (EFI_SIZE_TO_PAGES (Size));
+    if (mChipInfoCtxt.PartInfo) {
+      PartInfo = (PlatformInfoPartInfoType *)((UINTN *)Smem + Smem->PartInstancesOffset);
+      CopyMem (mChipInfoCtxt.PartInfo, PartInfo, Size);
+      mChipInfoCtxt.PartInfoLen = Smem->NumPartInstances;
+    }
+  }
+
```
leiflindholm, line 184: ...to here?

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Include/Library/PlatformInfoLib.h
```diff
@@ -0,0 +1,120 @@
...
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#pragma once
+
+#include <Base.h>
+#include <PlatformInfoDefs.h>
+
+/**
+  Return the platform type.
```
leiflindholm: These are exported functions, could you please give them proper descriptions with arguments and return values?

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Include/PlatformInfoDefs.h
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * Generate a platform version number.
+ *
+ * This macro generates the platform version number from the specified
+ * major and minor numbers. For example, version 1.2 is represented by
+ * @code PLATORMINFO_VERSION(1,2) @endcode
+ *
+*/
+#define PLATFORMINFO_VERSION(major, minor)       (((major) << 16) | (minor))
```
leiflindholm: Since it's `PlatformInfo`, it ought to be `PLATFORM_INFO".

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Include/PlatformInfoDefs.h
```diff
@@ -0,0 +1,273 @@
...
+  PLATFORMINFO_ERROR_OUT_OF_RANGE_PARAMETER  = -9,
+  PLATFORMINFO_ERROR_INVALID_ADDRESS         = -10,
+  PLATFORMINFO_ERROR_INSUFFICIENT_BUFFER_LEN = -11,
+} PlatformInfoResult;
+
+/**
+  Target type of the device on which the platform is running.
+ */
+typedef enum {
+  PLATFORMINFO_TYPE_UNKNOWN     = 0x00,                   /**< Unknown target device. */
```
leiflindholm: The comments on the names feel very "add 1 to i".

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Include/PlatformInfoDefs.h
```diff
@@ -0,0 +1,273 @@
...
+
+  PLATFORMINFO_KEY_32BITS = 0x7FFFFFFF
+} PlatformInfoKeyType;
+
+/**
+  Stores the target platform, the platform version, and the
+  platform subtype.
+ */
+typedef struct {
+  PlatformInfoPlatformType    PlatformType;  /**< Type of the current target. */
```
leiflindholm: "Add 1 to i."
That seems to be a pattern in this file. Can you prune some?

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Include/PlatformInfoDefs.h
```diff
@@ -0,0 +1,273 @@
...
+  /**< Build ID. */
+  UINT32                      RawChipId;             /**< Raw chip ID. */
+  UINT32                      RawChipVersion;        /**< Raw chip version. */
+  PlatformInfoPlatformType    PlatformType;          /**< Platform type. */
+  UINT32                      PlatformVersion;       /**< Platform version. */
+  UINT32                      Fusion;                /**< TRUE if Fusion; FALSE otherwise. */
+  UINT32                      PlatformSubtype;       /**< Platform subtype. */
+  PlatformInfoSmemPmicType    PmicInfo[PLATFORMINFO_SMEM_MAX_PMIC_DEVICES];
+
+  /**< DEPRECATED. May not contain the full
```
leiflindholm: I guess this refers to the above member array, but it sure doesn't look like it.
If it _is_ deprecated, can we rename it `Deprecated` with the comment "don't use"?

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Include/PlatformInfoDefs.h
```diff
@@ -0,0 +1,273 @@
...
+  UINT32                      PmicArrayOffset;       /**< Offset from base of structure to
+                                                        array of PlatformInfoSmemPmicType. */
+  UINT32                      ChipFamily;            /**< Chip family. */
+  UINT32                      RawDeviceFamily;       /**< Raw device family. */
+  UINT32                      RawDeviceNumber;       /**< Raw device number. */
+
+  UINT32                      QfpromChipId;                                     /**< QFPROM Chip ID   */
+  CHAR8                       ChipIdStr[PLATFORMINFO_SMEM_MAX_CHIP_ID_LENGTH];   /**< Chip name */
+  UINT32                      NumClusters;
+  /**< Number of clusters used by ChipInfo_GetDisabledCPUs API */
```
leiflindholm: Again with the comments on the line after the member. I guess this looks less bad in the dopxygen output, but it really messes with the source readability.
Are we sure the line length is a problem if we don't do the weird line break?
(more instances below)

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Library/PlatformInfoLib/PlatformInfo.c
```diff
@@ -0,0 +1,187 @@
...
+  SPDX-License-Identifier: BSD-2-Clause-Patent
+**/
+
+#include <Base.h>
+#include "PlatformInfoInternal.h"
+
+/*
+ * Driver context.
+ */
+PlatformInfoDrvCtxt  mPlatformInfoDrvCtxt;
```
leiflindholm: Can you do a global search/replace `Ctxt` -> `Ctx`?

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Library/PlatformInfoLib/PlatformInfoLib.c
```diff
@@ -0,0 +1,65 @@
...
+  if (DrvCtxt->Initialized) {
+    return PLATFORMINFO_SUCCESS;
+  }
+
+  Smem = SmemGetAddr (SmemHwSwBuildId, &Size);
+  if (Smem == NULL) {
+    return PLATFORMINFO_ERROR;
+  }
+
+  DrvCtxt->PlatformInfo.PlatformType = Smem->PlatformType;
```
leiflindholm: Can you make a helper function from here...

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Library/PlatformInfoLib/PlatformInfoLib.c
```diff
@@ -0,0 +1,65 @@
...
+      if (!DrvCtxt->Kvps) {
+        return PLATFORMINFO_ERROR;
+      }
+
+      Kvps = (PlatformInfoKvpsType *)((UINTN *)Smem + Smem->KvpsOffset);
+      CopyMem (DrvCtxt->Kvps, Kvps, Size);
+    }
+  }
+
+  DrvCtxt->Initialized = TRUE;
```
leiflindholm: ...to here?

Platform/Qualcomm/QualcommPlatformPkg/Library/ArmPlatformLib/ArmPlatformLibMem.c
```diff
@@ -20,15 +20,17 @@
...
 #include <Library/HobLib.h>
 #include <Library/PcdLib.h>
+#include <Library/PlatformInfoLib.h>
 #include <Library/RamPartitionTableLib.h>
 #include <Library/SmemLib.h>
 
-#include <PiPei.h>
-#include <Pi/PiHob.h>
 #include <Pi/PiBootMode.h>
+#include <Pi/PiHob.h>
```
leiflindholm: You know I like my sorted includes, but I like my functional/non-functional commit separation more.

leiflindholm (changes requested):

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/QualcommCommonSiliconPkg.dec
vishalo: For code that can be shared across MinPlatform and non MinPlatform.

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/QualcommCommonPlatformPkg.dec
vishalo: Let me create Readme.md with folder structure.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
vishalo: This is not LLM, this comes from downstream file, it's been there for ages ...  Will clean up description ...

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/QualcommCommonSiliconPkg.dec
b49020: QualcommSiliconPkg already plays that role, it has the SMEM library which is equally required on non MinPlatform.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
vishalo: @varunpopuri

varunpopuri: Right, this long predates LLMs.

> The description is also really confusing wrt difference between "raw version" and "version". Is this minor/major version?

The "raw" version is a HW revision counter that tracks all silicon revisions. It's an older way to track chip revisions, and is less practical than the major/minor chip version (see below) unless you really need flat versioning that ignores the major/minor distinction, or in the very rare case that a silicon change doesn't bump the major/minor version. Each chip has its own interpretation of these values: a chip that has a v1.0, v1.1, and v2.0 would have HW revisions 0, 1, and 2; whereas a chip that only has a v1.0 and v2.0 would have HW revisions 0 and 1. That means if you want to check for a board change, you need to know how many revisions the chip has had.

The version reported by `GetChipVersion` is the major/minor version, which are incremented for new silicon spins. Major versions typically correspond to full tapeouts, minor versions are for metal-layer revisions. These are the versions that SW cares about. There are very rare (~once a decade) cases that a silicon change doesn't get reflected as a major/minor version bump, in which case the raw version should be used. 

Perhaps "HW revision counter" would be better phrasing than "raw"? The "raw" terminology follows the convention in other functions, like `GetRawChipId` and `GetRawChipFamily`; it refers to information returned directly from HW without parsing.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
vishalo: @leiflindholm , Not LLM - I traced this back to 2016, and stopped checking further back ... regardless will fix .. :-)

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/ChipInfoDefs.h
vishalo: @leiflindholm  - your LLM sensor is likely due for recalibration. 

This code predates LLM .. :-) regardless, will fix it.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
vishalo: @leiflindholm : should we keep the raw and explain what it means?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
vishalo: @leiflindholm : are you asking to change the API to ChipInfoGetProductNameString (...) ?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
vishalo: Refers to SOC architecture family (architecture family spans across multiple SOC family), SOC family refers to different variants of SOC - Glymur variants for e.g ...

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
vishalo: /**
  Initialize the ChipInfo driver.

  If chip information can't be read (e.g., in a pre-silicon standalone
  build), each field is set to its documented UNKNOWN value instead of
  returning an error.

  @return  EFI_SUCCESS always.
**/

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
vishalo: /**
  Returns the HW part-level disabled feature details from fuses, read
  from the QTV/PTE region.

  @param[in]   Part  The ChipInfoPartType being queried.
  @param[in]   Idx   Hardware instance of the selected part.
  @param[out]  Mask  Used to store the part-specific details.

  @return  EFI_SUCCESS if successful.
  @return  EFI_INVALID_PARAMETER if Mask is NULL or Part is unknown.
  @return  EFI_INVALID_PARAMETER if Part is invalid.
  @return  EFI_NOT_READY if called before ChipInfoInit.
**/

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/ChipInfoDefs.h
vishalo: More than 4 defective cluster, and the part may never get to software teams and may be considered as screening failed.  @varunpopuri ?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Library/ChipInfoLib/ChipInfo.c
vishalo: <ChipInfoDefs.h> it is in include ...

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Include/Library/PlatformInfoLib.h
```diff
@@ -0,0 +1,120 @@
...
+  );
+
+/**
+  Return the OEM Variant.
+
+  OEM variants are OEM-specific customizations of a specific platform
+  type, subtype, and version.
+
+  Variant ID 0 => Qualcomm platform without any modifications.
+  IDs 1-255 are for OEM use.
```
varunpopuri: OEM variants can actually go up to 2^32-1, with CDT v6

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Include/PlatformInfoDefs.h
varunpopuri: Most of these acronyms have their full forms documented internally. We could use those as the comments. Or remove them altogether because whoever's using the device knows that they're using a QCB, for example, and it doesn't really matter what that means

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
varunpopuri: I don't think GetDisabledFeatures is even usable externally. The most that a caller would be able to do is check if _some_ component of an IP is disabled, but they wouldn't know what that component is. The meaning of each bit is only documented internally, and I'm not sure if we can publicize those. Until someone needs to use it, and until we publish the bit definition somewhere, I don't think we should expose this API.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
varunpopuri: It's supposed to be the name under which the chip is marketed, but even that isn't true anymore with the X Elite/Plus/... branding. This is a bad API that doesn't do what it claims, and it hasn't really been used for years. I'd rather not expose it externally unless we have to.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
varunpopuri: We already explain that the raw chip version is a HW revision counter where 0 is the initial tapeout. We use "raw" everywhere else (other functions, other images) to say that we're returning a value from HW without additional parsing. Giving the function a special name on UEFI doesn't help us in any way.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
vishalo: Removed from latest push.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
varunpopuri: @vishalo can you rephrase this as "The chip family is used to group all variants of an SOC"?

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/QualcommCommonSiliconPkg.dec
vishalo: Correct we need to move that to common as well.

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/QualcommCommonPlatformPkg.dec
vishalo: @b49020  https://github.com/tianocore/edk2-platforms/pull/1020/changes/6719e01bb01b6b7e157b2943c0fdcf7bc9e3cd6d#diff-740e12d1385815da4d7ad000e9d371e62ad320235cd082f0e596299da9c5db36

vishalo:
Updates based on review comments. Few pending more feedback.

vishalo:
@anushka-qcom

anushka-qcom:
Changes look appropriate.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/QualcommCommonSiliconPkg.dec
leiflindholm: OK, that's one of out 11 changed, please change also the remaining 10.

ricardosalveti:
Some findings with help from AI:

1. SMEM array offsets are scaled by 8 — all array reads are from the wrong address. In ChipInfoLib.c (ChipInfoPopulateCtxFromSmem) and PlatformInfoLib.c, every array is located with the pattern:
CpuClusters = (UINT32 *)((UINTN *)Smem + Smem->ClusterArrayOffset);
Casting to UINTN * before adding makes pointer arithmetic multiply the offset by 8 on AARCH64. These fields (ClusterArrayOffset, DisabledFeaturesArrayOffset, PartialFeaturesArrayOffset, CpuCoresArrayOffset, PartInstancesOffset, KvpsOffset) are byte offsets from the base of the socinfo structure — the same SMEM ABI the Linux kernel's drivers/soc/qcom/socinfo.c consumes with plain byte arithmetic, and the struct's own doc says "offset from base of structure". Every one of the six sites needs (UINT8 *)Smem + Offset. As written, the disabled-CPU masks, feature fuses, core counts, part instances, and KVPs are read from Smem + 8*offset — garbage at best, a fault past the SMEM region at worst. This also suggests the array paths were never validated against real socinfo data.

2. NULL dereference in ChipInfoGetNumCpuCores on socinfo formats 0.14–0.21. mChipInfoCtx.NumCores is only allocated when Format >= 0.22, but the API guards only with CpuCluster >= NumClusters, and NumClusters is populated from Format >= 0.14. On firmware reporting an in-between format, NumCores is NULL while NumClusters > 0, and the new logging loop in ArmPlatformLibMem.c calls this for every cluster during memory-map construction — a boot-time crash on older XBL. ChipInfoGetDisabledCpus has the same shape (CpuClusters can be NULL if AllocatePages failed, which is silently ignored during populate). Both need NULL checks.

3. Unclamped CopyMem into fixed-size globals. Smem->NumParts is used directly to size copies into DisabledFeatures[CHIPINFO_NUM_PARTS] and PartialFeatures[CHIPINFO_NUM_PARTS]. A newer XBL that knows more parts than this UEFI build overflows a global buffer. Clamp with MIN (Smem->NumParts, CHIPINFO_NUM_PARTS) — same defensive posture the RAM-partition code takes with RAM_NUM_PART_ENTRIES.

4. ChipInfoGetChipIdString can't deliver its documented contract. It returns EFI_SUCCESS unconditionally, but AsciiStrCpyS does not truncate — with MaxLength smaller than the stored string it fails (ASSERTs in DEBUG builds) and leaves the caller's buffer unwritten, while the header promises a "possibly truncated" copy. Either truncate explicitly (AsciiStrnCpyS-style) or propagate the error; also the NULL-pointer result from ChipInfoStrcpy is dropped. Related robustness nit: the init path copies Smem->ChipIdStr with plain AsciiStrCpyS, which fails (empty string plus DEBUG ASSERT) if the 32-byte SMEM field is fully packed without a terminator.

5. PlatformInfoGetDrvCtx latches init failure permanently. Its function-local STATIC Initialized flips to TRUE on the first call even when PlatformInfoInit fails (SMEM not ready), after which every accessor silently returns zeros forever unless someone calls PlatformInfoInit explicitly. The current call site initializes explicitly first, so it works today, but the double-flag dance (GetDrvCtx's static plus DrvCtx->Initialized) is fragile — gate the retry on DrvCtx->Initialized alone.

6. Layering inversion — the silicon package depends on the platform package. ChipInfoLib.inf (in Silicon/.../QualcommCommonSiliconPkg) lists QualcommCommonPlatformPkg.dec in [Packages] because the SMEM socinfo struct (PlatformInfoSmemType) lives in the platform package's PlatformInfoDefs.h. The socinfo layout is silicon-scoped (it sits beside SmemLib conceptually). Moving the SMEM struct definition into the silicon side removes the inversion — and would make the "Introduce ChipInfoLib" commit self-contained (today it references a DEC and header that only appear two commits later; nothing builds it until the final wiring commit, so bisect survives, but the commit can't compile in isolation). This adds to the unresolved question about why these aren't in QualcommSiliconPkg/QualcommPlatformPkg — worth settling before merge since #1020's whole premise is that those packages are being renamed to host exactly this kind of shared code.

7. INFs under-declare their dependencies. ChipInfoLib.inf uses AsciiStrCpyS (BaseLib) and AllocatePages (MemoryAllocationLib) but declares neither; PlatformInfoLib.inf uses MemoryAllocationLib and BaseMemoryLib but declares only SmemLib. They link today only because every consumer happens to pull those in. Also, the [Packages] paths Common/QualcommCommonPlatformPkg/... resolve through two different PACKAGES_PATH roots that each contain a Common/ directory — fully qualify them (Platform/Qualcomm/Common/..., Silicon/Qualcomm/Common/...) as the dsc.inc files already do.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/QualcommCommonSiliconPkg.dec
```diff
@@ -23,4 +23,4 @@
   Include
 
 [Guids]
-  gQualcommCommonSiliconPkgTokenSpaceGuid     = {0xDCC48FA6, 0X5910, 0X4A76, {0X90, 0X19, 0X7D, 0X67, 0X67, 0X21, 0XF3, 0XC2}}
+  gQualcommCommonSiliconPkgTokenSpaceGuid     = {0xDCC48FA6, 0x5910, 0x4A76, {0x90, 0x19, 0x7D, 0x67, 0x67, 0x21, 0xF3, 0xC2}}
```
leiflindholm, line 26: ...so squash this into previous please.

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/QualcommCommonPlatformPkg.dec
```diff
@@ -0,0 +1,27 @@
...
+# Include Section - list of Include Paths that are provided by this package.
+#                   Comments are used for Keywords and Module Types.
+#
+################################################################################
+
+[Includes.common]
+  Include
+
+[Guids]
+  gQualcommCommonPlatformPkgTokenSpaceGuid     = {0XB7BF47E4, 0X94FE, 0X4641, {0X83, 0X4F, 0XC5, 0XB9, 0X8E, 0XFF, 0XD5, 0XBE}}
```
leiflindholm: Here we go with uppercase Xs again.

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/QualcommCommonSiliconPkg.dec
leiflindholm: (OK, I see you accidentally split this fix across commits...)

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
leiflindholm: That does not make this any clearer.

OK, so it probably doesn't make sense to rename confusing terms that are consistently used internally, but let's please add some documentation to the file header.

That documentation should describe:
- what an architecture is
- what an architecture family is

it should describe that in ways that does not confuse this with regards to cpu architecture (e.g. ARMv8.2), SoC, and processor/cpu.

leiflindholm (changes requested):

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
leiflindholm: I think the updated text is OK.

vishalo:
> Some findings with help from AI:
> 
> 1. SMEM array offsets are scaled by 8 — all array reads are from the wrong address. In ChipInfoLib.c (ChipInfoPopulateCtxFromSmem) and PlatformInfoLib.c, every array is located with the pattern:
>    CpuClusters = (UINT32 *)((UINTN *)Smem + Smem->ClusterArrayOffset);
>    Casting to UINTN * before adding makes pointer arithmetic multiply the offset by 8 on AARCH64. These fields (ClusterArrayOffset, DisabledFeaturesArrayOffset, PartialFeaturesArrayOffset, CpuCoresArrayOffset, PartInstancesOffset, KvpsOffset) are byte offsets from the base of the socinfo structure — the same SMEM ABI the Linux kernel's drivers/soc/qcom/socinfo.c consumes with plain byte arithmetic, and the struct's own doc says "offset from base of structure". Every one of the six sites needs (UINT8 _)Smem + Offset. As written, the disabled-CPU masks, feature fuses, core counts, part instances, and KVPs are read from Smem + 8_offset — garbage at best, a fault past the SMEM region at worst. This also suggests the array paths were never validated against real socinfo data.
> 2. NULL dereference in ChipInfoGetNumCpuCores on socinfo formats 0.14–0.21. mChipInfoCtx.NumCores is only allocated when Format >= 0.22, but the API guards only with CpuCluster >= NumClusters, and NumClusters is populated from Format >= 0.14. On firmware reporting an in-between format, NumCores is NULL while NumClusters > 0, and the new logging loop in ArmPlatformLibMem.c calls this for every cluster during memory-map construction — a boot-time crash on older XBL. ChipInfoGetDisabledCpus has the same shape (CpuClusters can be NULL if AllocatePages failed, which is silently ignored during populate). Both need NULL checks.
> 3. Unclamped CopyMem into fixed-size globals. Smem->NumParts is used directly to size copies into DisabledFeatures[CHIPINFO_NUM_PARTS] and PartialFeatures[CHIPINFO_NUM_PARTS]. A newer XBL that knows more parts than this UEFI build overflows a global buffer. Clamp with MIN (Smem->NumParts, CHIPINFO_NUM_PARTS) — same defensive posture the RAM-partition code takes with RAM_NUM_PART_ENTRIES.
> 4. ChipInfoGetChipIdString can't deliver its documented contract. It returns EFI_SUCCESS unconditionally, but AsciiStrCpyS does not truncate — with MaxLength smaller than the stored string it fails (ASSERTs in DEBUG builds) and leaves the caller's buffer unwritten, while the header promises a "possibly truncated" copy. Either truncate explicitly (AsciiStrnCpyS-style) or propagate the error; also the NULL-pointer result from ChipInfoStrcpy is dropped. Related robustness nit: the init path copies Smem->ChipIdStr with plain AsciiStrCpyS, which fails (empty string plus DEBUG ASSERT) if the 32-byte SMEM field is fully packed without a terminator.
> 5. PlatformInfoGetDrvCtx latches init failure permanently. Its function-local STATIC Initialized flips to TRUE on the first call even when PlatformInfoInit fails (SMEM not ready), after which every accessor silently returns zeros forever unless someone calls PlatformInfoInit explicitly. The current call site initializes explicitly first, so it works today, but the double-flag dance (GetDrvCtx's static plus DrvCtx->Initialized) is fragile — gate the retry on DrvCtx->Initialized alone.
> 6. Layering inversion — the silicon package depends on the platform package. ChipInfoLib.inf (in Silicon/.../QualcommCommonSiliconPkg) lists QualcommCommonPlatformPkg.dec in [Packages] because the SMEM socinfo struct (PlatformInfoSmemType) lives in the platform package's PlatformInfoDefs.h. The socinfo layout is silicon-scoped (it sits beside SmemLib conceptually). Moving the SMEM struct definition into the silicon side removes the inversion — and would make the "Introduce ChipInfoLib" commit self-contained (today it references a DEC and header that only appear two commits later; nothing builds it until the final wiring commit, so bisect survives, but the commit can't compile in isolation). This adds to the unresolved question about why these aren't in QualcommSiliconPkg/QualcommPlatformPkg — worth settling before merge since [Platform/Qualcomm: Qualcomm MinPlatform Cleanup #1020](https://github.com/tianocore/edk2-platforms/pull/1020)'s whole premise is that those packages are being renamed to host exactly this kind of shared code.
> 7. INFs under-declare their dependencies. ChipInfoLib.inf uses AsciiStrCpyS (BaseLib) and AllocatePages (MemoryAllocationLib) but declares neither; PlatformInfoLib.inf uses MemoryAllocationLib and BaseMemoryLib but declares only SmemLib. They link today only because every consumer happens to pull those in. Also, the [Packages] paths Common/QualcommCommonPlatformPkg/... resolve through two different PACKAGES_PATH roots that each contain a Common/ directory — fully qualify them (Platform/Qualcomm/Common/..., Silicon/Qualcomm/Common/...) as the dsc.inc files already do.

Fixed these, Thanks @ricardosalveti  !

vishalo:
Thanks @leiflindholm . Please review - fixed .

Silicon/Qualcomm/Common/QualcommCommonSiliconPkg/Include/Library/ChipInfoLib.h
```diff
@@ -0,0 +1,373 @@
+/** @file
+  This is the interface specification for the chip driver/service.
+
+  An architecture, in the context of this file, is a hardware design
```
leiflindholm: maybe s/file/library/, but otherwise the description is good.

However, I think it looks confusing to put it above the copyright and license ... while (justifiably) verbose, this is basically a glossary entry.

vishalo: @leiflindholm - Thank you, fixed with latest push.

leiflindholm (approved):

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Library/PlatformInfoLib/PlatformInfoLib.c
```diff
@@ -0,0 +1,92 @@
...
+  UINT32                Size;
+  PlatformInfoKvpsType  *Kvps;
+
+  DrvCtx->PlatformInfo.PlatformType = Smem->PlatformType;
+  DrvCtx->PlatformInfo.Version      = Smem->PlatformVersion;
+  DrvCtx->PlatformInfo.Subtype      = Smem->PlatformSubtype;
+  DrvCtx->PlatformInfo.Fusion       = Smem->Fusion;
+
+  if (Smem->Format >= PLATFORM_INFO_VERSION (0, 18)) {
+    DrvCtx->NumKvps = Smem->NumKvps;
```
ricardosalveti: NumKvps gets set before calling AllocatePages, and it is also not set to 0 on failures during memory allocation, it just returns EFI_OUT_OF_RESOURCES.

Platform/Qualcomm/Common/QualcommCommonPlatformPkg/Library/PlatformInfoLib/PlatformInfoLib.c
```diff
@@ -0,0 +1,92 @@
...
+  IN     PlatformInfoSmemType  *Smem
+  )
+{
+  UINT32                Size;
+  PlatformInfoKvpsType  *Kvps;
+
+  DrvCtx->PlatformInfo.PlatformType = Smem->PlatformType;
+  DrvCtx->PlatformInfo.Version      = Smem->PlatformVersion;
+  DrvCtx->PlatformInfo.Subtype      = Smem->PlatformSubtype;
+  DrvCtx->PlatformInfo.Fusion       = Smem->Fusion;
```
ricardosalveti, line 38: What about OemVariantId, how is it getting populated?

vishalo:
@ricardosalveti Thank you, please see latest commit.

ricardosalveti (approved):

PR #1036, opened 2026-08-03
Silicon/Intel/Tools/FitGen: Set Rsvd Field to 0 For Optional Modules

### LINKED ISSUE #1035: [Bug]: FitGen.c Does Not Set Reserved Byte to 0 for Optional Modules - [Silicon/Intel/Tools]
### Is there an existing issue for this?

- [x] I have searched existing issues

### Bug Type

- [ ] Firmware
- [x] Tool
- [ ] Unit Test

### What packages are impacted?

Silicon/Intel/Tools

### Which targets are impacted by this bug?

RELEASE, DEBUG

### Current Behavior

Optional modules in fit gen (BPM, KM) do not have the reserved byte (byte 11) set to 0 as required in the spec: https://edc.intel.com/content/www/us/en/design/products-and-solutions/software-and-services/firmware-and-bios/firmware-interface-table/1.6/firmware-interface-table/. Lines 3250-3264

### Expected Behavior

As with all other entries in FitGen.c, this reserved byte should be set to 0 when building the entry.

### Steps To Reproduce

Build Intel reference BIOS and view FIT table before BPM Gen stitches in entries 

### Build Environment

```markdown
- OS(s): All supported
- Tool Chain(s): All supported
```

### Version Information

```text
Tip of main branch
```

### Urgency

Low

### Are you going to fix this?

I will fix it

### Do you need maintainer feedback?

No maintainer feedback needed

### Anything else?

_No response_

justynduthlerIntel:
fixes #1035. Reserved field in FIT entry for optional modules should be set to 0, except for CSE Secure Boot module, which uses the reserved field to store the subtype.

### DIFF
--- a/Silicon/Intel/Tools/FitGen/FitGen.c
+++ b/Silicon/Intel/Tools/FitGen/FitGen.c
@@ -3257,6 +3257,8 @@ FillFitTable (
     FitEntry[FitIndex].Type     = (UINT8)gFitTableContext.OptionalModule[Index].Type;
     if (FitEntry[FitIndex].Type == FIT_TABLE_TYPE_CSE_SECURE_BOOT) {
       FitEntry[FitIndex].Rsvd   = (UINT8)gFitTableContext.OptionalModule[Index].SubType;
+    } else {
+      FitEntry[FitIndex].Rsvd   = 0;
     }
     FitEntry[FitIndex].C_V      = 0;
     FitEntry[FitIndex].Checksum = 0;


Silicon/Intel/Tools/FitGen/FitGen.c
```diff
@@ -3257,6 +3257,8 @@ FillFitTable (
     FitEntry[FitIndex].Type     = (UINT8)gFitTableContext.OptionalModule[Index].Type;
     if (FitEntry[FitIndex].Type == FIT_TABLE_TYPE_CSE_SECURE_BOOT) {
       FitEntry[FitIndex].Rsvd   = (UINT8)gFitTableContext.OptionalModule[Index].SubType;
+    } else {
+      FitEntry[FitIndex].Rsvd   = 0;
```
SaiChaganty, line 3261: Wouldn't it be simpler to add this line after 3257 and remove the else condition all together ? :-)

justynduthlerIntel: Either or works. In the case of CSE secure boot then there would be 2 writes to this location. Using the existing conditional results in a single write no matter the entry. I can add it to after 3257 if that is preferred

SaiChaganty: let's leave it as you have it. I am not aware of major concern with single Vs dual writes. On a related note, there is a effort to make this whole tool modular. Perhaps we can do the cleanup there.

SaiChaganty (approved):

Silicon/Intel/Tools/FitGen/FitGen.c
justynduthlerIntel: Sounds good, more of a very very minor optimization that does not matter (conditional branching logic may make this a wash anyways)

ayushsriv79 (approved):

PR #1039, opened 2026-08-23
AmpereAltraPkg/Ac01PcieLib: pass Gen3 preset from NV correctly

sinuscosinustan:
The NVParam PCIE Gen3 Preset got ignored and instead the default Gen3 preset was always provided.


bexcran (approved):

bexcran:
This is dependent on https://github.com/tianocore/edk2-platforms/pull/926 so I'll merge this after that's in.

PR #1043, opened 2026-09-01
Platform/ARM: introduce TpmCrbFfaDeviceLib and rename FtpmSmm to TpmCrbFfaSmm

LeviYeoReum:
Currently, FtpmSmm is tied with TpmLib and this reduces a reusability of 
TPM CRB over FF-A generic handling code and prevent a TPM CRB over FF-A driver with dTPM.

Separate fTPM functionalities from FtpmSmm. For this, make a TpmCrbFfaDeviceLib and
implements this library with fTPM again. so that remain part of FtpmSmm which handles generic FF-A call handling
can be used and Let it be used with dTPM in the future.

Since fTPM feature is separated from FtpmSmm, rename it to TpmCrbOverFfaSmm.


Platform/ARM/Library/TpmCrbFfaFtpmLib/TpmCrbFfaFtpmLib.c
```diff
@@ -0,0 +1,1147 @@
...
+
+/**
+  This function dump TCG_EfiSpecIDEventStruct.
+
+  @param[in,out]  LogAddr    Event address
+  @param[in,out]  LogSize    Left size of events
+
+**/
+VOID
+DumpTcgEfiSpecIdEvent (
```
samimujawar: Should this function be made STATIC?

LeviYeoReum:
Gentle ping @samimujawar

Platform/ARM/Drivers/TpmCrbOverFfaSmm/TpmCrbOverFfaSmm.c
```diff
@@ -0,0 +1,426 @@
...
+  @param[in]  Status                  The EFI_STATUS code to be converted.
+  @param[in]  SuccessWithArgs         If success, return with result code.
+
+  @retval TPM2_FFA_SUCCESS_OK_RESULTS_RETURNED  If Status is EFI_SUCCESS and SuccessWithArgs == TRUE
+  @retval TPM2_FFA_SUCCESS_OK                   If Status is EFI_SUCCESS and SuccessWithArgs == FALSE
+  @retval others                                For all other status codes.
+**/
+STATIC
+TPM2_FFA_STATUS
+EfiStatusToTpm2FfaStatus (
```
amrathesh, line 48: nit: Is the omission of EFIAPI intentional, given that other static helpers include it?

Platform/ARM/Drivers/TpmCrbOverFfaSmm/TpmCrbOverFfaSmm.c
```diff
@@ -0,0 +1,426 @@
...
+  ARM_FFA_ARGS     *TpmArgs;
+  UINTN            Operation;
+  TPM2_FFA_STATUS  TpmStatus;
+
+  if ((CommBufferSize == NULL) || (*CommBufferSize < sizeof (ARM_FFA_ARGS))) {
+    DEBUG ((DEBUG_ERROR, "%a: Invalid Parameters\n", __func__));
+    return EFI_INVALID_PARAMETER;
+  }
+
+  TpmArgs   = CommBuffer;
```
amrathesh, line 345: Shouldn't we have a NULL check on CommBuffer before accessing it? 

and also, curious to know why CommBuffer and CommBufferSize marked optional even though this function can't work without it?

Platform/ARM/Drivers/TpmCrbOverFfaSmm/TpmCrbOverFfaSmm.c
```diff
@@ -0,0 +1,426 @@
+/** @file TpmCrbOverFfa StandaloneMm driver.
+
+  Copyright (c) 2024, Arm Limited. All rights reserved.<BR>
```
amrathesh: please update copyright year if new file.

Platform/ARM/Drivers/TpmCrbOverFfaSmm/TpmCrbOverFfaSmm.c
LeviYeoReum: > Shouldn't we have a NULL check on CommBuffer before accessing it?

This is totally redundant since the ArmPkg/Driver/StandaloneMmCpu always set the *Command buffer&.
And I'm also curious why I did it? anyway let me remove.

Thanks

Platform/ARM/Drivers/TpmCrbOverFfaSmm/TpmCrbOverFfaSmm.c
LeviYeoReum: Okay.

Platform/ARM/Drivers/TpmCrbOverFfaSmm/TpmCrbOverFfaSmm.c
LeviYeoReum: I missed. Thanks.

Platform/ARM/Library/TpmCrbFfaFtpmLib/TpmCrbFfaFtpmLib.c
```diff
@@ -0,0 +1,1150 @@
...
+        __func__,
+        Locality,
+        Status
+        ));
+      goto ErrorHandler;
+    }
+  }
+
+  Status = FtpmDeviceStartup ();
+  if (EFI_ERROR (Status)) {
```
sughoshg, line 1135: In the existing FtpmSmm.c module's entry point, the FtpmEventHandler() gets unregistered, if any of these functions return an error. However, with the current design, this de-registration of the handler is not done. Is that not an issue ?

Platform/ARM/Library/TpmCrbFfaFtpmLib/TpmCrbFfaFtpmLib.c
```diff
@@ -0,0 +1,1150 @@
...
+  )
+{
+  TPM2_COMMAND_HEADER   *Command;
+  TPM_CC                CommandCode;
+  UINT32                CommandSize;
+  TPM2_RESPONSE_HEADER  *Response;
+  TPM_RC                ResponseCode;
+  UINT32                ResponseSize;
+  PTP_CRB_REGISTERS     *CrbReg;
+
```
sughoshg, line 833: The logic in the existing FtpmSmm.c file is checking if `CurLocality != NextLocality` and returns an error if that is not the case. The check is missed int this implementation.

Platform/ARM/Library/TpmCrbFfaFtpmLib/TpmCrbFfaFtpmLib.c
```diff
@@ -0,0 +1,1150 @@
...
+STATIC INTERNAL_HASH_INFO  mHashInfo[] = {
+  { TPM_ALG_SHA256, HASH_ALG_SHA256, SHA256_DIGEST_SIZE, "SHA256" },
+  { TPM_ALG_SHA384, HASH_ALG_SHA384, SHA384_DIGEST_SIZE, "SHA384" },
+  { TPM_ALG_SHA512, HASH_ALG_SHA512, SHA512_DIGEST_SIZE, "SHA512" },
+};
+
+/**
+  Get Command Response Buffer address.
+
+  @param [in]   Locality          Locality
```
sughoshg, line 107: I think the function comments can be made a little more verbose. Also, there are some functions which are returning an error, but that is not being highlighted in the function comments, e.g. ValidateAndStripSpecIdEvent(). Kindly check the whole file for these issues.

Platform/ARM/Drivers/TpmCrbOverFfaSmm/Readme.md
sughoshg: This is not related to the above change, but since you are touching this file,

GCC5 -> GCC

sughoshg: GCC5 -> GCC

Platform/ARM/Library/TpmCrbFfaFtpmLib/TpmCrbFfaFtpmLib.c
LeviYeoReum: No, If the TpmCrbFfaFtpmLib's constructor is failed, the driver wouldn't be loaded (TpmCrbOverFfaSmm)
and the `TpmCrbOcerFfaSmm` wouldn't register *event handler* since its entry-point doesn't run.

sughoshg (commented):
In your commit message, the word "separated" is a little confusing imo. You can instead use something like "The fTPM functionality is moved to  TpmCrbFfaFtpmLib ..."

Platform/ARM/Library/TpmCrbFfaFtpmLib/TpmCrbFfaFtpmLib.c
LeviYeoReum: Thanks. and the check `Locality < CurLocality` is wrong (Should be inverse).
I'll fix it.

Platform/ARM/Drivers/TpmCrbOverFfaSmm/Readme.md
LeviYeoReum: Agree. Let me fix.

Platform/ARM/Library/TpmCrbFfaFtpmLib/TpmCrbFfaFtpmLib.c
sughoshg: > No, If the TpmCrbFfaFtpmLib's constructor is failed, the driver wouldn't be loaded (TpmCrbOverFfaSmm)
and the TpmCrbOcerFfaSmm wouldn't register event handler since its entry-point doesn't run.

I don't think that would be a guaranteed outcome. I believe the ProcessLibraryConstructorList() function returns a `void`, and the calls to the constructor functions are only checked with `ASSERT_EFI_ERROR()`. So, if a platform has opted out of a dead-loop on asserts, this will end up with the driver's entry point getting called ?

Platform/ARM/Drivers/TpmCrbOverFfaSmm/Readme.md
LeviYeoReum: Thanks. I'll fix it.

Platform/ARM/Library/TpmCrbFfaFtpmLib/TpmCrbFfaFtpmLib.c
LeviYeoReum: TBH, I'm not sure how locality could be more detail. locality itself is locality. why do we need additional comments for this? But. let me check other which doesn't specify return error.

Thanks.

Platform/ARM/Library/TpmCrbFfaFtpmLib/TpmCrbFfaFtpmLib.c
LeviYeoReum: You're right. I think it would be better to add interface whether to check Device initialised properly or not
So that, Driver to check it before registering handler.

Also, StandaloneMm couldn't support the notification. I'll remove useless interface too.

Thanks!

PR #1047, opened 2026-09-11
Drivers/OptionRomPkg/Bus/Usb/FtdiUsbSerialDxe: Fix control transfers, flow control node, and timeouts

mcassaniti:
## Description

This series addresses three issues in `Drivers/OptionRomPkg/Bus/Usb/FtdiUsbSerialDxe` that prevent the driver from operating correctly on compliant UEFI USB host controllers and with standard terminal drivers:

1. **Zero-length USB control transfers:** Vendor control requests (`SET_DATA`, `SET_BAUDRATE`, `SET_FLOW_CTRL`, `SET_DTR`, `SET_RTS`, `RESET_PORT`) pack all parameters in `wValue`/`wIndex` in the SETUP packet with `DevReq.Length = 0` (no USB DATA stage). The driver previously invoked `UsbControlTransfer()` specifying `EfiUsbDataOut`/`EfiUsbDataIn` with a 1-byte dummy stack variable and length 1, causing host controllers (such as `XhciDxe`/`EhciDxe`) to initiate an invalid DATA stage and return `EFI_DEVICE_ERROR`. This is aligned with Linux kernel prior art (`drivers/usb/serial/ftdi_sio.c` at tag `v6.13`).
2. **FlowControl Device Path:** The vendor flow control device path node was created with an uninitialized GUID (`00000000-0000-0000-0000-000000000000`) and appended unconditionally. `TerminalDxe` rejected the node, which led to no serial console binding. This patch initializes the GUID with `gEfiUartDevicePathGuid` and conditionally appends the node only when flow control is configured.
3. **Bulk transfer timeout:** Increases `FTDI_TIMEOUT` from 16ms to 50ms (3x the hardware latency timer) to provide margin for hypervisor scheduling jitter in virtualized environments (e.g. QEMU) without causing noticeable blocking latency on unresponsive endpoints.

## Testing

- Validated with `PatchCheck.py` across all commits (0 errors, 0 warnings).
- Compiled with GCC under standard EDK II BaseTools (`-Wall -Werror`).
- Verified in QEMU with `qemu-xhci` and emulated FTDI serial adapter with end-to-end bootloader and OS console.

Signed-off-by: Michael A Cassaniti <michael@cassaniti.id.au>


mcassaniti:
For reviewers interested in the background and testing context behind this fix series, take a look here:
https://github.com/mcassaniti/ftdi-uefi-serial

PR #1052, opened 2026-09-16
Platform/ARM/VExpressPkg: change memory region information when ENABLE_FIRMWARE_UPDATE is enabled

LeviYeoReum:
This patch includes the memory region change when ENABLE_FIRMWARE_UPDATE is enabled
according to change of TF-A[0] to utilise more memory in StandaloneMm and
some fixes for build and error of scripts.

Link: [0] https://review.trustedfirmware.org/c/TF-A/trusted-firmware-a/+/53044


Platform/ARM/VExpressPkg/ArmVExpress-FVP-AArch64.dsc
```diff
@@ -278,8 +282,16 @@
   # Normal pseudo crbs which locality from 0 to 3 are allocated
   # at the start of System Memory.
   #
+!if $(ENABLE_FIRMWARE_UPDATE) == TRUE
+  #
+  # When ENABLE_FIRMWARE_UPDATE, TZC DRAM size is 64MB.
+  #
+  gEfiSecurityPkgTokenSpaceGuid.PcdTpmBaseAddress|0xfbf10000
```
amrathesh, line 289: Seeing ^M in diff, could be some inconsistent line ending here.

LeviYeoReum: This file already is mixed with ^M and carriage return. and the GICv5's edk2-platform's patch series includes to change all with dos file.

So. this is ignorable right now. once GICv5's edk2-platform merge first, It will be rebased but
this doesn't hinder merging this patch if this is merged first.

